PR #48, opened 2019-08-06
WIP refactoring of recorder

ianb:
This moves the recorder into its own file, with no built-in UI.

This does not appear to ever stop the media stream, maybe because stm_vad isn't working.

Note the only UI for this is console messages, and it isn't wired to run any intents either.


ianb:
With the last commit this works. Nothing in popup.js is _great_, just kind of works, but it's ready to be wired up to the UI.

PR #52, opened 2019-08-08
Refactor the intent system

ianb:
- Move parsing into intentParser
- Move individual intents into extension/intents/
- Generate the list and put it in manifest.json
- Separate intents into a series of groups


extension/intents/playing/playing.js
```diff
@@ -0,0 +1,33 @@
...
+        url: googleQueryURL,
+      });
+    } else {
+      playerTab = await browser.tabs.getCurrent();
+    }
+
+    // get video content for the current tab
+    setTimeout(async () => {
+      await browser.tabs.executeScript(playerTab.id, {
+        file: "/extension/intents/playing/playMedia.js",
```
jcambre: I think the extension/ directory is the root path for everything, so this should be `file: "/intents/playing/playMedia.js",`

extension/intents/playing/playing.js
```diff
@@ -0,0 +1,33 @@
...
+      });
+      // TODO: poll for playing, instead of timeout
+    }, 3000);
+  });
+
+  this.intentRunner.registerIntent("pause", async (desc) => {
+    const currentTab = await browser.tabs.getCurrent();
+    // get video content for the current tab
+    const result = await browser.tabs.executeScript(currentTab.id, {
+      file: "/extension/intents/playing/pauseMedia.js",
```
jcambre: same, here: `file: "/intents/playing/pauseMedia.js",`

PR #117, opened 2019-08-15
Introduce search cards and within-service searches

jcambre:
In addition to the primary change of adding in search cards and within-service bang searches from DuckDuckGo, this PR also cleans up / removes a lot of assets that we're no longer using in the current version of Firefox Voice


extension/background/main.js
```diff
@@ -7,7 +7,10 @@ this.main = (function() {
       console.info("Executing intent:", desc);
       return intentRunner.runIntent(desc);
     }
-    console.error("Received message with unexpected type (${message.type}):", message);
+    console.error(
+      "Received message with unexpected type (${message.type}):",
```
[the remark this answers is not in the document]
ianb: Oops, yes, but I guess the diff is just from prettier here...

extension/background/intentParser.js
```diff
@@ -69,21 +43,36 @@ this.intentParser = (function() {
...
+    },
+    bangSearch: {
       matches: [
-        /^pause$/i,
+        /(?:do a )?(?:search (?:my |on |for )?|query |find(?: me)? |look up |lookup |look on |look for )(google slides|google docs|spotify|goodreads|mdn|coursera|google scholar|google drive|calendar|google calendar)(?: for (?:the )?)?(.*)/i,
       ],
+      slots: ["service", "query"],
     },
-    read: {
+    bangSearchAlt: {
```
ianb: I created #119 to remove the necessity for two intents here.

extension/intents/playing/playing.js
```diff
@@ -10,7 +10,7 @@ this.intents.playing = (function() {
         url: googleQueryURL,
       });
     } else {
-      playerTab = (await browser.tabs.query({active: true}))[0];
+      playerTab = (await browser.tabs.query({ active: true }))[0];
```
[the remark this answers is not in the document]
ianb: I think we are? There's a state where Firefox is open with no windows, but then nothing really happens. I plan to make this `desc` thing into an object with handlers for many cases like this, probably including an active tab getter, that at least will be a good place to put a guard on something like this.

extension/manifest.json.ejs
```diff
@@ -43,12 +43,13 @@
     "storage",
     "tabs",
     "mozillaAddons",
-    "about:reader*"
+    "about:reader*",
+    "https://api.duckduckgo.com/*"
```
ianb: Do we need this? Probably not, but it's not a big deal either

extension/popup/ui.js
```diff
@@ -154,6 +154,15 @@ this.ui = (function() {
     });
   };
 
+  exports.showCard = function showCard(data) {
+    document.querySelector("#popup").classList.add("hasCard");
+    document.querySelector("#card-header").innerText = data.Heading;
```
ianb: `.textContent` would be the better thing to set here (innerText is a little quirky, mostly good as a getter)

extension/popup/popup.html
```diff
@@ -37,16 +37,30 @@
...
+            <div id="card-summary"></div>
+            <div id="card-ack">
+              <div id="ddg-ack">
+                <img id="ddg-logo" src="./images/dax-logo.svg"/><a href="https://duckduckgo.com/">Results from DuckDuckGo</a>
+              </div>
+              <div class="sep"></div>
+              <div id="source-ack">
+                <span>Source:</span>
+                <div id="card-source">
+                  <a id="card-source-link"></a>
```
ianb: Do we need a `target=_blank` on this? I'm honestly not sure, I can't remember what the link behavior is here.

extension/popup/ui.js
```diff
@@ -178,14 +187,18 @@ this.ui = (function() {
     settingsIcon.addEventListener("click", showSettings);
   }
 
-  function closePopup() {
+  exports.closePopup = function closePopup(ms = 2500) {
```
ianb: What's the reason for the timeout here? Enough time for someone to read the result before closing? The time should at least get a name and comment at the top of the file.

extension/popup/vad.js
```diff
@@ -156,6 +156,9 @@ this.vad = (function() {
       log.debug("goCloud, why:", why);
       this.stopGum();
       // FIXME: maybe we need to signal the UI here?
+      if (why === "GoCloud finishedvoice") {
+        ui.setState("processing"); // REVIEW: not sure if this should send a message to popup.js or emit an event some other way?
```
ianb: Technically the "right" way is to send the message to voice.js, and then to popup.js, and then to ui.js. I don't even know if this has a handle to voice.js (or the `Recorder` instance) at the moment. Maybe best to just have a TODO here.

extension/searching.js
```diff
@@ -12,6 +12,49 @@ this.searching = (function() {
     return searchUrl.href;
   };
 
+  exports.ddgEntitySearch = async function(query) {
+    const response = await fetch(
+      `https://api.duckduckgo.com/?q=${query}&format=json&pretty=1&skip_disambig=1`
```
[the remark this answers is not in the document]
ianb: Yes! `encodeURIComponent(query)`

extension/searching.js
```diff
@@ -12,6 +12,49 @@ this.searching = (function() {
...
+    );
+    const ddgData = await response.json();
+    if (!ddgData.AbstractText) {
+      // the response from DDG was null, and there was no matching Instant Answer result
+      return {};
```
ianb: Should this return null?

extension/searching.js
```diff
@@ -12,6 +12,49 @@ this.searching = (function() {
...
+    }
+    const cardData = (({
```
ianb: I don't understand what this is doing.

extension/searching.js
```diff
@@ -12,6 +12,49 @@ this.searching = (function() {
...
+      "google calendar": "gcal",
+      "google drive": "drive",
+      spotify: "spotify",
+      goodreads: "goodreads",
+      mdn: "mdn",
+      coursera: "coursera",
+    };
+    const bang = SERVICE_BANG_MAP[service.toLowerCase()];
+    const response = await fetch(
+      `https://api.duckduckgo.com/?q=!${bang}+${query}&format=json&pretty=1&no_redirect=1`
```
ianb: Ditto quoting here. Even `bang` should be quoted, just so we can verify on this line alone that the query string can't be malformed.

ianb:
Some notes, but we should move it forward today to get it landed with followups as necessary.

extension/searching.js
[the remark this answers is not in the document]
jcambre: Is it best to do `encodeURIComponent(query)` or generate the URL w/ `new URL...` and `url.searchParams.set(...)`?

extension/popup/popup.html
```diff
@@ -73,8 +87,10 @@
         <div id="settings-icon">
             <svg xmlns="http://www.w3.org/2000/svg" width="16" height="16" viewBox="0 0 16 16"><path fill="context-fill" d="M15 7h-2.1a4.967 4.967 0 0 0-.732-1.753l1.49-1.49a1 1 0 0 0-1.414-1.414l-1.49 1.49A4.968 4.968 0 0 0 9 3.1V1a1 1 0 0 0-2 0v2.1a4.968 4.968 0 0 0-1.753.732l-1.49-1.49a1 1 0 0 0-1.414 1.415l1.49 1.49A4.967 4.967 0 0 0 3.1 7H1a1 1 0 0 0 0 2h2.1a4.968 4.968 0 0 0 .737 1.763c-.014.013-.032.017-.045.03l-1.45 1.45a1 1 0 1 0 1.414 1.414l1.45-1.45c.013-.013.018-.031.03-.045A4.968 4.968 0 0 0 7 12.9V15a1 1 0 0 0 2 0v-2.1a4.968 4.968 0 0 0 1.753-.732l1.49 1.49a1 1 0 0 0 1.414-1.414l-1.49-1.49A4.967 4.967 0 0 0 12.9 9H15a1 1 0 0 0 0-2zM5 8a3 3 0 1 1 3 3 3 3 0 0 1-3-3z"></path></svg>
         </div>
-        <div id="logo">
+        <div id="moz-voice-privacy">
+          <a href="">How Mozilla ensures voice privacy</a>
         </div>
+        <div></div>
```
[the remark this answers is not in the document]
jcambre: A bit of a hack because the parent container is a flexbox... I was trying to find an easy way of placing the settings icon in the bottom left and the "How Mozilla ensures voice privacy" link in the bottom center.

extension/popup/ui.js
jcambre: Oops! Sorry.

extension/popup/popup.html
jcambre: Oh, good call. I hadn't tested that. It looks like by default, and with target="_blank", it opens in a new window, even though my browser preference is to open links in new tabs. I imagine our ideal behavior is new tab within the same window, right?

extension/popup/ui.js
jcambre: Yes, just to give people a bit of time to glance at it before it disappears. I'll make it a const

extension/popup/vad.js
jcambre: 👍

extension/searching.js
jcambre: https://stackoverflow.com/questions/17781472/how-to-get-a-subset-of-a-javascript-objects-properties

The JSON that we get from DDG's API can be huge, with lots of keys we don't actually need for the card, so I wanted to just pick out the ones that actually get used within the UI. If that's unnecessary, I'm happy to cut this and pass along the full json? Or can add a comment on what it's doing.

extension/searching.js
jcambre: Ah, yep! Sorry. Did a dumb thing to make the linter happy.

extension/popup/popup.html
ianb: Ideally, yes... I vaguely recall using `browser.tabs.create()` to do this in the past because it's the only way I could get it to work in a new tab.

extension/searching.js
ianb: That's fine, but yeah, add a comment to clarify

extension/searching.js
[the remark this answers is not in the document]
ianb: Either way is fine, though just using `encodeURIComponent` is less mental overhead.

extension/popup/popup.html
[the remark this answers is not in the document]
jcambre: LMK if I should do this some other way... Will otherwise leave it in for now

extension/searching.js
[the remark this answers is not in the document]
jcambre: 👍 I've updated it to use `encodeURIComponent` both here and in `ddgBangSearchUrl`

extension/background/intentParser.js
jcambre: will add a TODO w/ a reference to that issue in the code

extension/searching.js
jcambre: Done

extension/popup/popup.html
jcambre: Ok, changed this to use `browser.tabs.create()`. Listening for clicks on that link also means we can dismiss the popup after navigating to that new tab, which is handy.

extension/searching.js
jcambre: Done

pdehaan:
`+342 −13,461` may be one of the greatest PRs I've ever seen. 👏

PR #124, opened 2019-08-16
Add stylelint

pdehaan:


.stylelintrc
```diff
@@ -0,0 +1,8 @@
+{
+  "extends": "stylelint-config-standard",
+  "ignoreFiles": "extension/css/vendor/*.css",
+  "rules": {
+    "font-family-no-missing-generic-family-keyword": null,
```
pdehaan: This "fixed" the following error:

```
extension/popup/popup.css
 15:16  ✖  Unexpected missing generic font family   font-family-no-missing-generic-family-keyword
```

This seems to be coming from the following rule, which I'm guessing doesn't like that there isn't a fallback font:

```css
body,
.cls-1 {
  font-family: "Inter" !important;
}
```

🤷‍♂

extension/popup/popup.css
```diff
@@ -105,7 +105,7 @@ body,
 }
 
 #header-title {
-  font-size: 1.125rrem;
+  font-size: 1.125rem;
```
pdehaan: Previously reported as `Unexpected unknown unit "rrem"`:

```
extension/popup/popup.css
 108:14  ✖  Unexpected unknown unit "rrem"           unit-no-unknown
```

PR #147, opened 2019-08-28
Loading Fonts Locally

### LINKED ISSUE #123: Load fonts locally
I believe Inter is currently getting loaded remotely: https://github.com/mozilla/firefox-voice/blob/755f3eff681786b2d4cef9bc5164772176c7be83/extension/popup/popup.css#L1

We should shift to using local fonts so the popup can run fully offline.

Patil2099:
Fixes #123


ianb:
Thanks for the patch @Patil2099 !  Can you change the PR to include just the three font files that @awallin mentioned?

Patil2099:
@ianb Are the changes correct?
Thanks!

ianb:
Looks great, thanks @Patil2099 !

PR #337, opened 2019-09-27
Add tab open via voice.

acwio:
Simple support for opening a new tab. Could be improved to include a specific target (i.e. URL) or a predetermined number of tabs.


awallin:
In addition, we should also support open new window. 

Any chance a modifier of open private tab/window can be included?

ianb:
Another simple variation (no reason for it to be part of this PR) would be `open [query] in new tab` which would open a tab and use a google I'm-lucky search as the URL (seeing `searching.js`).

Issue #112, opened 2019-08-14
[browser command] Add print intent


ianb:
It would open up the print dialog.

It's easy because [it's already an API](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/tabs/print)

awallin:
Be nice if you could say "Print" a second time to execute from the dialog box.

KhushilMistry:
Hey, I want to work on this issue. I will submit a PR ASAP.

ianb:
Closed in #474

PR #474, opened 2019-10-23
#112 [browser command] Add print intent

KhushilMistry:
#112 [browser command] Add print intent.


ianb:
Nice, you're our contributor to add an intent!

Issue #491, opened 2019-10-24
Convert popup.html and ui.js to React


ianb:
I have a hard time handling the UI in the popup, and probably the time has come to modernize it.

A conservative React setup is my own comfort spot.

PR #507, opened 2019-10-26
Initial commit with react and most popup functionality

jenniferharmon:
PR for Issue #491 

Hi! I stumbled across this issue while browsing for Hacktoberfest.

This Pull Request copies React libs from `node_modules/` to the `js/vendor/` directory. It currently uses the minified production version of React, but can be switched to use the development version.

`popup.js` has been added to .gitignore because it is now auto-generated from `popup.jsx` using babel.

I moved the `ui.js` logic into `popup.jsx`. I noticed vad.js calls ui.setState(). I would be happy to update that to use browser.runtime.sendMessage() and create a new type for changing the popup state.

Do you have any suggestions for how I might improve the Pull Request?


extension/popup/popup.css
```diff
@@ -15,91 +15,13 @@ body,
   font-family: "Inter" !important;
 }
 
-.listening #voice-input {
```
ianb: Removing CSS is great! I haven't been doing a great job of that myself :-/

extension/popup/popup.jsx
```diff
@@ -0,0 +1,744 @@
+/* globals util, voice, vad, lottie, settings, log, voiceShim, buildSettings */
+
+const React = window.React
```
ianb: If you put `React` and `ReactDOM` in the `/* globals */` section then these assignments won't be necessary.

extension/popup/popup.jsx
```diff
@@ -0,0 +1,744 @@
...
+        }
+        if (
+            !backgroundTabRecorder &&
+            isWaitingForPermission &&
+            Date.now() - isWaitingForPermission < FAST_PERMISSION_CLOSE
+        ) {
+            this.startOnboarding()
+        }
+
+        // TODO: offload mic and other resources before closing?
```
ianb: Does this all get run even when the window is closed? I don't think so, and so I think most of this cleanup should happen in a `beforeunload` handler. (Probably this component will never unmount?)

extension/popup/popup.jsx
```diff
@@ -0,0 +1,744 @@
...
+    updateExamples = async () => {
+        const suggestions = await browser.runtime.sendMessage({
+            type: "getExamples",
+            number: 3,
+        })
+
+        this.setState({ suggestions })
+    }
+
+    setCurrentState = (currentState) => {
```
ianb: I'm not sure this is the right pattern to be using with React, or at least not in its entirety. I think it makes sense for a couple cases:

1. Waiting for voice
2. Waiting for typing (once typing has begun)
3. After a command is heard/entered.

The third case has a bunch of cases, like a card, an error message, some text, etc. But those can each be properties, and each displayed if the data is present.

That said, it's probably best to leave this as it is for now without trying to change everything on the first pass. This is already a big patch!

jenniferharmon:
Thanks for the great feedback @ianb 

I will try to address these issues in the next day or two.

jenniferharmon:
I believe my latest commit makes it safe to resolve the merge conflicts by accepting my changes to those files. I ported the logic in b5a81bacf6f142ae414a493b5e917540 to popup.jsx. 

I was not able to trigger a scenario where a search card gets updated. Any suggestions for testing that?

ianb:
If the changes are applied, you should be able to merge master and then `git add` the deleted files to resolve the conflict.

For a card, search for "weather".

jenniferharmon:
Hi @ianb I think I have addressed all of your suggestions, with the exception of the unload event handling, which I believe was also not working before the refactor. Please let me know if there is anything else you would like done or changed to accept the PR.

ianb:
Thanks, this is great! There might be some details to deal with, but I think it's easier to do that in-tree given the size of the patch. So we'll get it landed and take it from there.

I can tell you've figured out quite a bit about the project on your own here, I hope we'll be seeing more contributions from you!

jenniferharmon:
Thanks! This was my first time working on a browser extension. I had fun and learned a lot about the project while working on the refactor. Feel free to `@mention` me on any issues you think I might be well-suited to work on.

PR #548, opened 2019-11-07
Use babel source-maps for more precise debugging

jenniferharmon:
Those changes look great @ianb ! The dev functionality to open a tab on start is awesome. Thanks for cc'ing me.

Have you considered using babel's `--source-map` instead of `--retain-lines`? It seems like this would provide more accurate line information for debugging.


ianb:
I've avoided sourcemaps in the past because they don't work everywhere in an extension (see [Bug 1401241](https://bugzilla.mozilla.org/show_bug.cgi?id=1401241)), but they'll work in the popup and other pages, since these are "normal" pages (`moz-extension:` aside), so sure, let's do that!

PR #568, opened 2019-11-13
translation added

maloney8:
Translation to different languages

### DIFF
--- a/extension/background/languages.js
+++ b/extension/background/languages.js
@@ -1,7 +1,7 @@
 this.languages = (function() {
     const exports = {};
 
-    exports.languageCode = {
+    const languageCodes = {
         Czech: "cs",
         Danish: "da",
         Dutch: "nl",
@@ -24,7 +24,9 @@ this.languages = (function() {
         Ukrainian: "uk",
     };
 
-    exports.getLanguages = function() {
-        return exports;
+    exports.languageCode = function() {
+        return Object.keys(languageCodes);
     };
+
+    return exports;
 })();
--- a/extension/background/languages.js
+++ b/extension/background/languages.js
@@ -24,7 +24,7 @@ this.languages = (function() {
         Ukrainian: "uk",
     };
 
-    exports.languageCode = function() {
+    exports.languageNames = function() {
         return Object.keys(languageCodes);
     };
 
--- a/extension/background/languages.js
+++ b/extension/background/languages.js
@@ -1,9 +1,7 @@
-this.languages = {};
-
-this.languageList = (function() {
+this.languages = (function() {
     const exports = {};
 
-    const LANGUAGES = {
+    exports.languageCode = {
         Czech: "cs",
         Danish: "da",
         Dutch: "nl",
@@ -26,7 +24,7 @@ this.languageList = (function() {
         Ukrainian: "uk",
     };
 
-    exports.allLanguages = function() {
-        return Object.keys(LANGUAGES);
+    exports.getLanguages = function() {
+        return exports;
     };
 })();
--- a/extension/background/languages.js
+++ b/extension/background/languages.js
@@ -1,0 +1,32 @@
+this.languages = {};
+
+this.languageList = (function() {
+    const exports = {};
+
+    const LANGUAGES = {
+        Czech: "cs",
+        Danish: "da",
+        Dutch: "nl",
+        English: "en",
+        Finnish: "fi",
+        French: "fr",
+        German: "de",
+        Hungarian: "hu",
+        Italian: "it",
+        Norwegian: "no",
+        Polish: "pl",
+        Portuguese: "pt",
+        Romanian: "ro",
+        Russian: "ru",
+        Slovak: "sk",
+        Slovenian: "sl",
+        Spanish: "es",
+        Swedish: "sv",
+        Turkish: "tr",
+        Ukrainian: "uk",
+    };
+
+    exports.allLanguages = function() {
+        return Object.keys(LANGUAGES);
+    };
+})();
--- a/extension/intents/navigation/navigation.js
+++ b/extension/intents/navigation/navigation.js
@@ -56,18 +56,22 @@ this.intents.navigation = (function() {
 
   this.intentRunner.registerIntent({
     name: "navigation.translate",
-    description: "Translate the given page to the chosen language, using Google Translate",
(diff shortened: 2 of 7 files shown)


extension/background/languages.js
```diff
@@ -0,0 +1,32 @@
+this.languages = {};
+
+this.languageList = (function() {
```
ianb: Each file should just export a variable that matches the name of the file. So it should be `this.languages = (function() ...)`

extension/background/languages.js
```diff
@@ -0,0 +1,32 @@
...
+        Spanish: "es",
+        Swedish: "sv",
+        Turkish: "tr",
+        Ukrainian: "uk",
+    };
+
+    exports.allLanguages = function() {
+        return Object.keys(LANGUAGES);
+    };
+})();
```
ianb: This needs `return exports` at the end, that's your basic problem.

extension/intents/navigation/navigation.js
```diff
@@ -56,14 +56,16 @@ this.intents.navigation = (function() {
 
   this.intentRunner.registerIntent({
     name: "navigation.translate",
-    description: "Translate the given page to English, using Google Translate",
+    description: "Translate the given page to the chosen language, using Google Translate",
     match: `
-    translate (this |) (page | tab | article | site |) (to english |) (for me |)
```
ianb: This pattern should stay in, and default to English (since the tool only understands English, this is an appropriate default).

extension/intents/navigation/navigation.js
```diff
@@ -56,14 +56,16 @@ this.intents.navigation = (function() {
...
+    translate (this |) (page | tab | article | site |) to [language:lang]
     `,
-    examples: ["Translate this page", "test:translate"],
+    examples: ["test:translate this page to Spanish", 
+    "test:translate this page to Dutch"],
     async run(context) {
+      const language = context.slots.language;
```
ianb: And then this should default to English, i.e., `const language = context.slots.language || "en"`

extension/background/languages.js
```diff
@@ -0,0 +1,32 @@
+this.languages = {};
+
+this.languageList = (function() {
+    const exports = {};
+
+    const LANGUAGES = {
```
ianb: I'd export this like `exports.languageCodes = {...}`

extension/intents/navigation/navigation.js
```diff
@@ -56,14 +56,16 @@ this.intents.navigation = (function() {
...
       const tab = await context.activeTab();
-      const translation = `https://translate.google.com/translate?hl=&sl=auto&tl=en&u=${encodeURIComponent(
+      const translation = `https://translate.google.com/translate?hl=&sl=auto&tl=${language}&u=${encodeURIComponent(
```
ianb: You have to map `language` to the language code, like `${languages.languageCode[language]}`

extension/background/languages.js
```diff
@@ -0,0 +1,32 @@
+this.languages = {};
```
ianb: Looks like this variable isn't needed anywhere

ianb:
BTW, make sure to run `npm run format` and `npm test` as well

extension/intents/navigation/navigation.js
maloney8: Fixed.

extension/intents/navigation/navigation.js
maloney8: Fixed.

extension/intents/navigation/navigation.js
maloney8: Fixed.

extension/background/languages.js
maloney8: Fixed.

extension/background/languages.js
maloney8: Fixed.

extension/background/languages.js
maloney8: I fixed the most of it, but I am still not sure how should my languages.js file look like, since this way it's not working.

```
this.languages = (function() {
    const exports = {};

    exports.languageCode = {
        Czech: "cs",
        Danish: "da",
        Dutch: "nl",
        English: "en",
        Finnish: "fi",
        French: "fr",
        German: "de",
        Hungarian: "hu",
        Italian: "it",
        Norwegian: "no",
        Polish: "pl",
        Portuguese: "pt",
        Romanian: "ro",
        Russian: "ru",
        Slovak: "sk",
        Slovenian: "sl",
        Spanish: "es",
        Swedish: "sv",
        Turkish: "tr",
        Ukrainian: "uk",
    };

    exports.getLanguages = function() {
        return exports;
    };
})();
```

Or maybe I am calling it the wrong way in the **intentParser.js** :
```
const ENTITY_TYPES = {
    serviceName: serviceList.allServiceNames(),
    musicServiceName: serviceList.musicServiceNames(),
    lang: languages.getLanguages(),
  };
```

ianb: Remember the whole module is a function, and `this.languages` is the result of that function – so you need `return exports;` right at the end of the function. As it is the function isn't returning anything, and so `this.languages` is undefined.

maloney8: Alright, I think it's finally working.
Screenshot_1

One more question. Shall I give this functionality to the `navigation.translateSelection` intent too?

ianb: Yes, all this makes sense for translateSelection as well

extension/background/languages.js
```diff
@@ -0,0 +1,32 @@
...
+        Russian: "ru",
+        Slovak: "sk",
+        Slovenian: "sl",
+        Spanish: "es",
+        Swedish: "sv",
+        Turkish: "tr",
+        Ukrainian: "uk",
+    };
+
+    exports.languageCode = function() {
```
ianb: A bit of naming confusion here: these are the language names (e.g., ,"Italian") not the codes ("it").

maloney8: Fixed.

extension/background/languages.js
maloney8: Okay, done. Are we finished?

ianb:
This looks good. I realized my suggestion about `npm run format` caught some files it shouldn't, but I'll fix those back up after landing

PR #612, opened 2019-11-25
Update lexicon.html.ejs

awallin:
Added pin tab example to lexicon

### DIFF
--- a/extension/views/lexicon.html.ejs
+++ b/extension/views/lexicon.html.ejs
@@ -204,6 +204,7 @@
           <div class="utterance">
             Example
             <p>Close tab</p>
+            <p>Pin tab</p>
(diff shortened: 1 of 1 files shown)


maitrella:
@awallin For each new intent should we update lexicon file?.

ianb:
Yeah, maybe we should add a note to the intent writing document. I thought about generating the lexicon from the code, but creating something with a reasonable structure isn't very easy.

maitrella:
I'll keep in mind for future intents. Thanks.

awallin:
Not every intent will need to be included in the Lexicon. We'll have to make a case by case decision. For example the "joke" intent is more of an easter egg, while the pin tab is relevant functionality and thus inculded in the Lexicon.

Issue #356, opened 2019-10-02
[Settings] Configuring timeout for listening.


acwio:
It'd be great to be able to configure how quickly the listener times out. Often times, I'll know what I want, and I'll speak my words quite quickly. Other times, I'll want to take my time in getting my words out.

ianb:
Here's the code that we'd be parameterizing: https://github.com/mozilla/firefox-voice/blob/68d53f0eb0ac9e31bf3aae28139c58f448d4bb95/extension/popup/vad.js#L42-L47

ianb:
I'm now thinking we probably don't want or need this to be configurable, we should just choose a good value. I've increased it to 15 seconds in #613, though we could adjust that further. (@awallin ?)

awallin:
Agreed, at this time we don't need the configuration. 15 seconds is a good place to start, based on user feedback, our own usage, and observations during user testing we can adjust.

PR #613, opened 2019-11-25
Increase initial silence timeout to 15 seconds

ianb:
Relates to #356


awallin:
@ianb this only affects the initial timeout, but won't cause the timeout after speaking has begun to be 15 seconds, correct?

ianb:
Correct, 1.5 seconds of silence still defines the end.

PR #635, opened 2019-12-03
Onboarding redesign

jenniferharmon:
- [x] Convert onboarding view to React
- [x] Save user setting for transcript opt-in/out
- [x] Add page elements and styles per designs
- [ ] Add final approved copy text
- [x] Add Footer
- [ ] Add Responsive layout
- [x] Only show transcript opt-in/out once
- [ ] Polish and make styles pixel perfect (time permitting)


.gitignore
```diff
@@ -14,6 +14,9 @@ extension/popup/*.js.map
 extension/options/optionsController.js
 extension/options/optionsView.js
 extension/options/*.js.map
+extension/onboarding/onboardingController.js
+extension/onboarding/onboardingView.js
+extension/onboarding/*.js.map
```
ianb: We can just add `*.js.map` (with no path) to catch all of these

extension/onboarding/onboardingController.jsx
```diff
@@ -0,0 +1,69 @@
...
+      if (location.pathname.endsWith("onboard.html")) {
+        launchPermission();
+      }
+      const result = await settings.getSettingsAndOptions();
+      userSettings = result.settings;
+    };
+
+    const setOptinValue = async value => {
+      // TODO: Is this the right setting to update?
+      userSettings.collectAudio = value;
```
ianb: This should be `userSettings.utterancesTelemetry`

extension/onboarding/onboardingController.jsx
```diff
@@ -0,0 +1,69 @@
...
+      await settings.saveSettings(userSettings);
+    };
+
+    const launchPermission = async () => {
```
ianb: This should only be called if they've already opted in or out (so we aren't giving them the permission dialog at the same time as the telemetry dialog)

So something like:

```
if (optinViewAlreadyShown) {
  launchPermission();
}
```

extension/onboarding/onboardingController.jsx
```diff
@@ -0,0 +1,69 @@
...
+      try {
```
ianb: Potentially right here could be `setPermissionError("Waiting")` (or something like that, where we say that we're waiting for the user to enable the microphone)

extension/onboarding/onboardingController.jsx
```diff
@@ -0,0 +1,69 @@
...
+        // TODO (in view): Set hotkey suggestion based on navigator
+        // document.querySelector("#action-key").textContent =
+        //     navigator.platform === "MacIntel" ? "Command ⌘" : "Ctrl";
+        const tracks = stream.getTracks();
+        for (const track of tracks) {
+          track.stop();
+        }
+      } catch (e) {
+        if (e.name === "NotAllowedError") {
+          setPermissionError("NotAllowedError");
```
ianb: This error means the user specifically clicked "Don't Allow"

extension/onboarding/onboardingController.jsx
```diff
@@ -0,0 +1,69 @@
...
+        } else {
+          setPermissionError("UnknownError");
```
ianb: This is some other unexpected error, like no microphone present. We might want to include `String(e)`

jenniferharmon:
@ianb There are some minor details not yet complete (responsive design and possibly some copy text changes), but I think this is is pretty good shape if you want to review it and consider merging.

extension/onboarding/onboard.html
```diff
@@ -4,212 +4,14 @@
     <meta charset="UTF-8" />
     <title>Welcome to Firefox Voice</title>
     <link rel="stylesheet" href="onboarding.css" />
-    <link rel="stylesheet" href="../css/vendor/tailwind.min.css" />
```
ianb: I think this file is now unused and can be deleted

extension/onboarding/onboard.html
```diff
@@ -4,212 +4,14 @@
...
-      </div>
-    </div>
-
-    <script src="onboarding.js"></script>
+    <div id="onboarding-container"></div>
+    <script src="../js/vendor/react.production.min.js"></script>
+    <script src="../js/vendor/react-dom.production.min.js"></script>
+    <script src="../settings.js"></script>
+    <script src="onboardingView.js"></script>
+    <script src="onboardingController.js"></script>
```
ianb: This should have these additional scripts, which handle some error catching:

```html
    <script src="../buildSettings.js"></script>
    <script src="../log.js"></script>
    <script src="../catcherAsyncSetup.js"></script>
    <script src="../catcher.js"></script>
```

extension/onboarding/onboardingController.jsx
```diff
@@ -0,0 +1,67 @@
...
+      userSettings = await settings.getSettings();
+      setOptinViewShown(userSettings.collectTranscriptsOptinShown);
+
+      if (optinViewAlreadyShown) {
+        launchPermission();
+      }
+    };
+
+    const setOptinValue = async value => {
+      userSettings.collectTranscriptsOptinShown = true;
```
ianb: I think "answered" might be a better term than "shown" (since it can show more than once). Also this should be set to `Date.now()` (to record the date) instead of just a boolean.

ianb:
The comments are all things that should get followed up on soon, as well as #651, but we'll land. Lots of good work!

extension/onboarding/onboardingController.jsx
ianb: I've fixed this in  #661

Issue #618, opened 2019-11-27
Remove "for internal use only"


ianb:

awallin:
Known locations:
- Onboarding page
- Bottom of the doorhanger

maitrella:
Should the "div" sections be completely removed or just the text?

ianb:
Probably removed? I'd remove them, see if it breaks the layout, and if not then they should be removed. If it does break the layout, then they should not be removed ;)

maitrella:
Ok. I also removed "During the internal pilot..." because I think is related to "internal use" concept.
Correct me if I'm wrong.

PR #636, opened 2019-12-03
Remove "for internal use only"

maitrella:
Remove "for internal use only" #618

### DIFF
--- a/extension/popup/popupView.jsx
+++ b/extension/popup/popupView.jsx
@@ -286,10 +286,6 @@ this.popupView = (function() {
             ></path>
           </svg>
         </div>
-        <div id="moz-voice-privacy">
-          <strong>For Mozilla internal use only</strong>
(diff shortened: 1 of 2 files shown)


extension/popup/popupView.jsx
```diff
@@ -286,10 +286,6 @@ this.popupView = (function() {
             ></path>
           </svg>
         </div>
-        <div id="moz-voice-privacy">
-          <strong>For Mozilla internal use only</strong>
-          {/* <a href="">How Mozilla ensures voice privacy</a> */}
```
ianb: @awallin do we want to restore that privacy link? I forgot that I overwrote it with the internal use message.

awallin: Because our FAQ needs work let's leave the link out for now. I'll open a new issue for the FAQ and when that is in better shape we'll add the link.

PR #639, opened 2019-12-03
Fix navigation.translateSelection intent

### LINKED ISSUE #623: "Translate" commands display "Internal error: TypeError: languages.languageCode is undefined" in "Doorhanger"
### Prerequisistes:
Mic permissions are enabled.
A page in a different language than English is opened and active, like https://ro.wikipedia.org/wiki/Leu.

### STR:
2. Say or write a "Translate" command, e.g. "Translate page to English".

### Expected result:
The page is translated in English without issues.

### Actual result:
An "Internal error: TypeError: languages.languageCode is undefined" error is displayed in the "Doorhanger" and the page is not translated.

### Notes:
Reproduced also for the "Translate selection" command. (2nd gif)

maitrella:
Fix #623 navigation.translateSelection intent

### DIFF
--- a/extension/intents/navigation/navigation.js
+++ b/extension/intents/navigation/navigation.js
@@ -99,7 +99,7 @@ this.intents.navigation = (function() {
         throw e;
       }
       const url = `https://translate.google.com/#view=home&op=translate&sl=auto&tl=${
-        languages.languageCode[language]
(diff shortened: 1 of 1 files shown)


ianb:
Oh, I missed that one, thanks!

PR #649, opened 2019-12-04
First draft of previous result intent

### LINKED ISSUE #562: "Previous result" command makes a new Google search
### Prerequisites:
Mic permissions are enabled.
A "Search lion" command is made to have the result in the "Doorhanger" and two "Next result" commands are made consecutively to have the new tab opened with the second website from the Google search list.

### STR:
2. Say or write a "Previous result" command.

### Expected result:
The previous website is displayed in the tab regarding lions.

### Actual result:
A new Google search is made for "previous result" and the first website from the search list opens in a new tab.

### Notes:

maitrella:
I need a lot of feedback XD
Trying to fix #562


extension/intents/search/search.js
```diff
@@ -244,6 +244,56 @@ this.intents.search = (function() {
...
 
+  intentRunner.registerIntent({
+    name: "search.previous",
+    description:
+      "If you've done a search then this will display the previous search result.",
+    examples: ["test:previous search item"],
+    match: `
+    (search |) previous (search |) (result{s} | item{s} | page | article |)
+    `,
+    async run(context) {
```
ianb: Probably `.run()` should call something like `moveResult(-1)` (and then "next" would call `moveResult(1)`) – since most of this code is the same

extension/popup/popupView.jsx
```diff
@@ -558,6 +559,15 @@ this.popupView = (function() {
...
           ) : null}
+          {previous ? (
+            <div id="search-show-previous">
+              Say <strong>previous result</strong> to view: <br />
+              <strong id="search-show-previous-title">{previous.title}</strong>
+              <span id="search-show-previous-domain">
+                {new URL(previous.url).hostname}
+              </span>
+            </div>
+          ) : null}
```
ianb: We probably don't have to show anything about the previous result – the user already saw it, and should have some familiarity with it. We might want to suggest that "previous" is an option to say, but for now I think we should just leave it out of the view entirely.

extension/intents/search/search.js
maitrella: Thank you for the feedback. After my holidays I,ll continue working on it.

extension/popup/popupView.jsx
maitrella: Ok. :+1:

maitrella:
@ianb I created the moveResult function to avoid duplicated code. Sorry if I done something wrong with my first rebase (I need feedback on this point also).

ianb:
@maitrella it looks like things got really mixed up with the rebasing. Are all your changes in `search.js`? If so, I might recommend taking a copy of that file and reapplying it to master.

maitrella:
Correct me if i'm wrong I have to close this pool request and open a new one based on master only with changes in search.js?.

ianb:
You could force-push a clean branch over this branch, but probably the easiest thing is to just close this one and do another separate PR.

maitrella:
I can’t thank you enough !

Issue #679, opened 2019-12-06
Command examples from the bottom of the "Welcome to Firefox Voice" page are not centered aligned with the rest of the page text


alexandra-martin:
### Prerequisites:
Mic permission are enabled.
"Allow" or "Don't Allow" button from “Allow Firefox Voice to Collect Voice Transcriptions” pop-up needs to be clicked.

### STR:
1. Observe the "Welcome to Firefox Voice" page.

### Expected result:
Page doesn't have display issues.

### Actual result:
The command examples from the bottom part of the page are not center aligned with the rest of the text from the page.

### Notes:
Reproduced on Mac 10.14.6 and Win10x64 with Firefox Nightly 72.0a1 (64-bit).
"Go to New York Times" and the rest of the examples have the length 1027 px, while "Say things like" has the length 1067 px.

![ov3](https://user-images.githubusercontent.com/54401091/70308118-450ac380-1813-11ea-9df7-a99691387c30.gif)

alexandra-martin:
Verified on Mac 10.14.6 and Win10x64 with Firefox Nightly 72.0a1 (64-bit).

![1](https://user-images.githubusercontent.com/54401091/70519111-173fba80-1b44-11ea-8a17-986db125dc38.gif)

PR #682, opened 2019-12-06
Adds toolbar arrow to onboarding page.

jenniferharmon:
I think this is a pretty good tablet and desktop (css only) solution for getting the arrow pointing from the get started section to the extension toolbar. We'll need to do something a bit different for mobile. I think if this solution doesn't meet expectations, I'll have to look into a javascript solution.

Also fixes bug #679


awallin:
Mobile is a lower priority. Doesn't have to be figure out by Monday's beta.

PR #717, opened 2019-12-10
Makes search 'next' item clickable

### LINKED ISSUE #709: Next search item should be clickable
The item at the bottom:

image

it's just sitting there asking to be clicked, but it isn't clickable.

jenniferharmon:
Fixes #709

### DIFF
--- a/extension/popup/popupView.jsx
+++ b/extension/popup/popupView.jsx
@@ -580,7 +580,7 @@ this.popupView = (function() {
         {next ? (
           <a
             id="search-show-next"
-            href={new URL(next.url)}
+            href={next.url}
             target="_blank"
             rel="noopener"
           >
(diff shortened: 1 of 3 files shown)


extension/popup/popupView.jsx
```diff
@@ -578,7 +578,12 @@ this.popupView = (function() {
           ) : null}
         </div>
         {next ? (
-          <div id="search-show-next">
+          <a
+            id="search-show-next"
+            href={new URL(next.url)}
```
ianb: Why `{new URL(...)}`? Couldn't this just be `next.url`?

But we should probably make this a bit more complex, and it should do the same thing that saying "next" does. Maybe this is fine:

```js
        browser.runtime.sendMessage({
          type: "runIntent",
          "next",
        });
```

At least that's a fine starting point.

PR #722, opened 2019-12-11
Open new window intent

### LINKED ISSUE #695: [Feature] Open a new window / private window
Now you can say "Open a new tab", but also "Open a new window" (or private browsing) would be great feature 😄

maitrella:
Fix #695 but need feedback on private window intent.


ianb:
Yeah, the private window thing is a probably from it being turned off generally in the extension: https://github.com/mozilla/firefox-voice/blob/master/extension/manifest.json.ejs#L22

I'll open a separate ticket about that.

PR #735, opened 2019-12-12
Shows minimized or expanded listening view based on voice interaction…

### LINKED ISSUE #324: Doorhanger behavior
Implement the following doorhanger behavior to create an efficient UX:

* After trigger the doorhanger opens to a "Listen" state showing only a title, the Zap, settings icon, and link to privacy statement.

* After 3 seconds of no audio activity the doorhanger expands to reveal example utterances and feedback for previous requests.

* Note: The first 5 uses of Firefox Voice after installation should override the above behavior and show the fully expanded doorhanger to suggest things that can be said.

* When timeout occurs (6 seconds?) the doorhanger should auto close itself (instead of displaying an error).

doorhanger-behavior

jenniferharmon:
… and number of visits

Fixes #324 

- [x] After trigger the doorhanger opens to a "Listen" state showing only a title, the Zap, settings icon, and link to privacy statement.
- [x] After 3 seconds of no audio activity the doorhanger expands to reveal example utterances and feedback for previous requests.
- [x] The first 5 uses of Firefox Voice after installation should override the above behavior and show the fully expanded doorhanger to suggest things that can be said.
- [ ] When timeout occurs (6 seconds?) the doorhanger should auto close itself (instead of displaying an error).

### DIFF
--- a/extension/popup/popup.css
+++ b/extension/popup/popup.css
@@ -132,6 +132,8 @@ body,
   opacity: 0;
   margin: 0;
   height: 0;
+  position: absolute;
+  top: 0;
 }
 
 #search-results {
--- a/extension/popup/popupController.jsx
+++ b/extension/popup/popupController.jsx
@@ -25,6 +25,7 @@ this.popupController = (function() {
     const [searchResult, setSearchResult] = useState(null);
     const [cardImage, setCardImage] = useState(null);
     const [recorderVolume, setRecorderVolume] = useState(null);
+    const [expandListeningView, setExpandedListeningView] = useState(true);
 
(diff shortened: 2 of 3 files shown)


jenniferharmon:
@ianb I am not sure how to implement the last item in the list.

extension/popup/popup.css
```diff
@@ -132,6 +132,8 @@ body,
   opacity: 0;
   margin: 0;
   height: 0;
+  position: absolute;
```
jenniferharmon: Fixes a scrolling issue in the listening view caused by the hidden text input.

PR #748, opened 2019-12-13
Uses one state for userSettings in optionsController

### LINKED ISSUE #738: Use one state for userSettings in optionsController
Right now each option has its own state/setter and update function. We could simplify things by just putting these all together, with one state (that matches userSettings) and one update function.

jenniferharmon:
Fixes #738


ianb:
Many deleted lines, very nice!

Issue #753, opened 2019-12-13
Add "duplicate tab" intent


maitrella:
When someone asks for:

    Duplicate this tab/page

Expected result: Duplicate the current tab/page in the same window.

PR #754, opened 2019-12-13
Duplicate current tab intent

maitrella:
Issue #753


maitrella:
@awallin Should I update the lexicon with this too?.

awallin:
Yes, it's unique. Add it under 'Browser control.'

PR #768, opened 2019-12-13
Added possible voice command for the read intent

vanekcsi:
I added three non-mandatory options for the read voice intent keeping the existing functionality, extending it with a couple of options.


ianb:
Thanks, extending the language we match is very helpful!

PR #789, opened 2019-12-17
Zoom in, out, reset intents

maitrella:
First draft with default Firefox zoom factors. I need a lot of feedback on this.

### DIFF
--- a/extension/intents/tabs/tabs.js
+++ b/extension/intents/tabs/tabs.js
@@ -1,4 +1,86 @@
 this.intents.tabs = (function() {
+  const MAX_ZOOM = 3;
+  const MIN_ZOOM = 0.3;
+  const DEFAULT_ZOOM = 1;
+  const defaultZoomValues = [
+    0.3,
+    0.5,
+    0.67,
+    0.8,
+    0.9,
+    1,
+    1.1,
+    1.2,
+    1.33,
+    1.5,
+    1.7,
+    2,
+    2.4,
+    3,
+  ];
+
+  async function updateZoom(context, operation) {
+    // 0 -> Reset
+    // 1 -> zoom in
+    // -1 -> zoom out
+    const activeTab = await context.activeTab();
+    if (operation === 0) {
+      await browser.tabs.setZoom(activeTab.id, DEFAULT_ZOOM);
+    } else {
+      const zoomFactor = await browser.tabs.getZoom(activeTab.id);
+      if (
+        (zoomFactor >= MAX_ZOOM && operation === 1) ||
+        (zoomFactor <= MIN_ZOOM && operation === -1)
+      ) {
+        const e = new Error("Zoom limit reached");
+        e.displayMessage = "Zoom limit reached";
+        throw e;
+      }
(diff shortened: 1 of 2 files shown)


maitrella:
Fix #790

extension/intents/tabs/tabs.js
```diff
@@ -1,4 +1,86 @@
...
+      const zoomFactor = await browser.tabs.getZoom(activeTab.id);
+      if (
+        (zoomFactor >= MAX_ZOOM && operation === 1) ||
+        (zoomFactor <= MIN_ZOOM && operation === -1)
+      ) {
+        const e = new Error("Zoom limit reached");
+        e.displayMessage = "Zoom limit reached";
+        throw e;
+      }
+      const pos = defaultZoomValues.indexOf(zoomFactor);
```
ianb: Since the user can update the zoom level in other ways, we can't assume the zoom level will be in defaultZoomValues (e.g., `zoomFactor` might be 2.5). There's a couple ways to handle it, I suppose – maybe loop through the list (maybe construct a backwards list with `Array.from(defaultZoomLevels).reverse()`) and find the next value.

extension/intents/tabs/tabs.js
```diff
@@ -1,4 +1,86 @@
...
+      );
+    }
+  }
+
+  this.intentRunner.registerIntent({
+    name: "tabs.zoomIn",
+    description: "Zoom in current tab",
+    examples: ["Zoom in tab"],
+    match: `
+    zoom in (this |) (tab |) (for me |)
```
ianb: I think even `in` could be optional here, since zooming usually implies zooming in.

@awallin do you have other ideas for phrases? Increase/decrease size?

extension/intents/tabs/tabs.js
```diff
@@ -1,4 +1,86 @@
...
+    },
+  });
+
+  this.intentRunner.registerIntent({
+    name: "tabs.zoomReset",
+    description: "Reset zoom in current tab",
+    examples: ["Reset zoom"],
+    match: `
+    reset zoom (for me |)
+    zoom reset (for me |)
```
ianb: `reset size` definitely makes sense here

maitrella: Ok

extension/intents/tabs/tabs.js
maitrella: Thanks for the feedback. I've tried to implement the same behaviour as Ctrl+ & Ctrl- as a proof of concept.

What do you think about using toolkit.zoomManager.zoomValues as defaultZoomValues?. Maybe is a more general solution for zoom intents.

Upss and zoom.maxPercent setting?. The more I read, the less I know how to code this group of intents. :)

ianb: Well, it can definitely get too complicated. I thought it could just go like this:

```js
function nextLevel(levels, current) {
  for (const level of levels) {
    if (current < level) {
      return level;
    }
  }
  return levels[levels.length - 1];
}
```

And then just another similar function for zooming down.

ianb:
Looks great, thanks!

PR #804, opened 2019-12-18
Prevent onboarding page from being opened multiple times

espertus:
While this code keeps the onboarding page from being opened in multiple tabs, it does not properly switch to an existing tab, despite the call to `browserUtil.makeTabActive()`. My hope is that a reviewer could let me know what I'm doing wrong. Thank you.

### DIFF
--- a/extension/popup/popupController.jsx
+++ b/extension/popup/popupController.jsx
@@ -50,18 +50,14 @@ this.popupController = (function() {
     const init = async () => {
       const userSettings = await settings.getSettings();
       if (!userSettings.collectTranscriptsOptinAnswered) {
-        console.log("Searching for open onboarding tabs...");
+        let onboardingURL = browser.runtime.getURL("/onboarding/onboard.html");
         for (const tab of await browser.tabs.query({
-          url: ["moz-extension://*/onboarding/onboard.html"]
+          url: [onboardingURL]
         })) {
-          console.log("Found an open onboarding tab!");
-          window.close();
-          console.log("Trying to open tab " + tab.id);
-          await browserUtil.makeTabActive(tab);
-          return;
+          await browser.tabs.remove(tab.id);
         }
         browser.tabs.create({
-          url: browser.runtime.getURL("onboarding/onboard.html"),
+          url: onboardingURL,
         });
         window.close();
         return;
--- a/extension/popup/popupController.jsx
+++ b/extension/popup/popupController.jsx
@@ -50,6 +50,16 @@ this.popupController = (function() {
     const init = async () => {
       const userSettings = await settings.getSettings();
       if (!userSettings.collectTranscriptsOptinAnswered) {
+        console.log("Searching for open onboarding tabs...");
+        for (const tab of await browser.tabs.query({
+          url: ["moz-extension://*/onboarding/onboard.html"]
+        })) {
+          console.log("Found an open onboarding tab!");
+          window.close();
+          console.log("Trying to open tab " + tab.id);
+          await browserUtil.makeTabActive(tab);
+          return;
+        }
         browser.tabs.create({
           url: browser.runtime.getURL("onboarding/onboard.html"),
         });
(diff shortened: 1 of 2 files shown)


extension/popup/popupController.jsx
```diff
@@ -50,6 +50,16 @@ this.popupController = (function() {
...
       const userSettings = await settings.getSettings();
       if (!userSettings.collectTranscriptsOptinAnswered) {
+        console.log("Searching for open onboarding tabs...");
+        for (const tab of await browser.tabs.query({
+          url: ["moz-extension://*/onboarding/onboard.html"]
+        })) {
+          console.log("Found an open onboarding tab!");
+          window.close();
+          console.log("Trying to open tab " + tab.id);
+          await browserUtil.makeTabActive(tab);
```
ianb: I'm not confident that this works from the popup – honestly I've never figured out the pattern of what works where. But it's possible it has to be run from the background.

extension/popup/popupController.jsx
```diff
@@ -50,6 +50,16 @@ this.popupController = (function() {
     const init = async () => {
       const userSettings = await settings.getSettings();
       if (!userSettings.collectTranscriptsOptinAnswered) {
+        console.log("Searching for open onboarding tabs...");
+        for (const tab of await browser.tabs.query({
+          url: ["moz-extension://*/onboarding/onboard.html"]
+        })) {
+          console.log("Found an open onboarding tab!");
+          window.close();
```
ianb: This should happen at the end, it probably stops everything. In fact it might be necessary to do `setTimeout(window.close)` just to make sure other things have time to run before the popup closes.

extension/popup/popupController.jsx
```diff
@@ -50,6 +50,16 @@ this.popupController = (function() {
     const init = async () => {
       const userSettings = await settings.getSettings();
       if (!userSettings.collectTranscriptsOptinAnswered) {
+        console.log("Searching for open onboarding tabs...");
+        for (const tab of await browser.tabs.query({
+          url: ["moz-extension://*/onboarding/onboard.html"]
```
ianb: Note you can also use `browser.runtime.getURL("/onboarding/onboard.html")` to get the exact URL.

espertus:
Thanks for the quick and helpful feedback, @ianb. 

I have a solution that deletes any existing onboarding tab before creating a new one. Let me know if this approach is acceptable, in which case I will generalize it to the other pages, or if I should try a different approach, such as communicating a request to a background script. 

Suggestions always welcome!

ianb:
@espertus I tried using `await browserUtil.makeTabActive(tab);` (instead of remove) and it seemed to work fine. Though if you missed using `await` then probably the window would have closed before that function could finish.

Anyway, switching to use makeTabActive should work.

Also browserUtil should either be used, or not included in popup.html. And if you run `npm test` it will do some basic linting tests (one of those tests is why the check on the PR isn't successful).

espertus:
> @espertus I tried using `await browserUtil.makeTabActive(tab);` (instead of remove) and it seemed to work fine. Though if you missed using `await` then probably the window would have closed before that function could finish.

As shown in our earlier conversation (which I just now unresolved), I originally had `await browserUtil.makeTabActive(tab);`

I'll try it out again.

> Also browserUtil should either be used, or not included in popup.html. 

Absolutely right. I forgot to remove it.

> And if you run `npm test` it will do some basic linting tests (one of those tests is why the check on the PR isn't successful). Also `npm run format` (or if you configure your editor, I have mine setup to run prettier on every save).

Thank you for these tips.

espertus:
> As shown in our earlier conversation (which I just now unresolved), I originally had `await browserUtil.makeTabActive(tab);`
> 
> I'll try it out again.

Okay @ianb, `await browserUtil.makeTabActive(tab)` worked once I moved `window.close()` below it, as you had suggested. Thank you.

I'll close this draft PR, generalize the solution, run tests, than submit a real PR.

PR #805, opened 2019-12-18
Go to homepage intent

### LINKED ISSUE #747: Support 'go to my homepage' intent
Based on telemetry users are trying to go to their home page using voice.

'Go to my homepage' or 'Open my homepage' should navigate existing tab to the homepage as defined in about:preferences.

maitrella:
Fix #747

### DIFF
--- a/extension/intents/tabs/tabs.js
+++ b/extension/intents/tabs/tabs.js
@@ -154,9 +154,10 @@ this.intents.tabs = (function() {
     `,
     async run(context) {
       const result = await browser.browserSettings.homepageOverride.get({});
-      result.value
-        .split("|")
-        .map(homepageUrl => browser.tabs.create({ url: homepageUrl }));
+      const homePageUrls = result.value.split("|");
+      for (const homePageUrl of homePageUrls) {
+        await browser.tabs.create({ url: homePageUrl });
+      }
     },
   });
 })();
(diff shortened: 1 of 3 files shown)


extension/intents/tabs/tabs.js
```diff
@@ -144,4 +144,19 @@ this.intents.tabs = (function() {
...
+    description: "Opens HomePage in a tab",
+    examples: ["open my homepage"],
+    match: `
+    (open | go to |) (my |) homepage (for me|)
+    `,
+    async run(context) {
+      const result = await browser.browserSettings.homepageOverride.get({});
+      result.value
+        .split("|")
+        .map(homepageUrl => browser.tabs.create({ url: homepageUrl }));
```
ianb: I'm not sure when or why there can be multiple URLs here, so I'm not sure if it's better to open the first URL or open all the URLs. I guess opening them all is an interesting choice at least? Maybe whoever somehow has multiple URLs will then be inspired to give us feedback.

It's not a big concern, but `.forEach()` would be more appropriate here than `.map()`, since it doesn't make use of the return values. But, one issue here is that any errors won't be caught – each `browser.tabs.create()` returns a promise, but nothing watches the promise. You could technically do `await Promise.all(result.value...)`, or just make this a simple loop with awaits.

maitrella: You can put as many URLs as you want in homepage preferences in Firefox that's why I implemented this way. Thanks for the feedback and thanks for sharing your knowledge.

Issue #701, opened 2019-12-10
New tab opens every time the mic icon is clicked and permissions are not given


alexandra-martin:
### Prerequisites:
Add-on is installed for the first time.
Mic permissions are not enabled.

### STR:
1. Click on the mic icon several times.

### Expected result:
A single "Welcome to Firefox Voice" tab is displayed in the browser.

### Actual result:
Several "Welcome to Firefox Voice" tabs open.

### Notes:
Reproduced on Mac 10.14.6 and Win10x64 with Firefox Nightly 72.0a1 (64-bit).
Reproduced also if mic permissions are enabled and “Allow Firefox Voice to Collect Voice Transcripts” pop-up is displayed. (2nd gif)

![4](https://user-images.githubusercontent.com/54401091/70539900-fb9cda00-1b6c-11ea-9d28-2f64307acf04.gif)

![6](https://user-images.githubusercontent.com/54401091/70539915-035c7e80-1b6d-11ea-8f5b-737a4393cae3.gif)

ianb:
The fix we'd like is instead of opening new pages, it should activate the existing onboarding page (and maybe reload it to make sure that it's state is up-to-date).

alexandra-martin:
#### Update:
The issue is reproduced also after clicking on the "Things you can say" and "Privacy" links from the "Lexicon" and "Privacy Policy" pages.

![1](https://user-images.githubusercontent.com/54401091/70606622-8fba8000-1c05-11ea-9951-c40bb47f73c4.gif)

![2](https://user-images.githubusercontent.com/54401091/70606504-4833f400-1c05-11ea-81ad-6f4eefe29b07.gif)

espertus:
May I try fixing this issue?

ianb:
@espertus You got it! This is an example of a guarded tab open: https://github.com/mozilla/firefox-voice/blob/master/extension/intents/search/search.js#L22-L55

espertus:
I think this should be closed.

PR #815, opened 2019-12-19
Reduce loading of onboarding, lexicon, and privacy pages.

espertus:
This addresses issue #701.

### DIFF
--- a/extension/browserUtil.js
+++ b/extension/browserUtil.js
@@ -33,6 +33,28 @@ this.browserUtil = (function() {
     });
   };
 
+  exports.activateTab = async function activateTab(url) {
+    if (!url.includes("://")) {
+      url = browser.runtime.getURL(url);
+    }
+    for (const tab of await browser.tabs.query({
+      url: [url]
+    })) {
+      browserUtil.makeTabActive(tab);
+      return;
+    }
+    browser.tabs.create({
+      url
+    });
+  };
+
+  exports.activateTabClickHandler = async function activateTabClickHandler(event) {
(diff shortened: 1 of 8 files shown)


espertus:
This PR successfully prevents the onboarding page from being opened multiple times.

This PR removes redundant opens of `lexicon.html` in `onboard.html`, `options.html`,
and `popup.html` but not `self.js`, `lexicon.html.ejs`, and `privacy-policy.html.js`.

This is not ready for merging, since it does not eliminate all redundant opens of `lexicon.html` and does not reduce loads of `privacy-policy.html`. 

I created a PR to get feedback before making further changes.

extension/browserUtil.js
```diff
@@ -33,6 +33,28 @@ this.browserUtil = (function() {
...
+    if (!url.includes("://")) {
+      url = browser.runtime.getURL(url);
+    }
+    for (const tab of await browser.tabs.query({
+      url: [url]
+    })) {
+      browserUtil.makeTabActive(tab);
+      return;
+    }
+    browser.tabs.create({
```
ianb: both browserUtil and browser.tabs calls should return (both to give the caller the tab object if they want, and to send back errors if they have an exception)

ianb:
I'll land it as-is, most of those other links aren't something people should encounter too often, and opening multiple tabs is something that happens often in the browser anyway. `self.js` could use it, and already has `browserUtil` loaded (it runs in the background page, where everything listed in `manifest.json.ejs` gets loaded).

PR #819, opened 2019-12-19
Further reduce duplicated tabs

espertus:
This further addresses issue #701 and completes my planned changes (unless a reviewer suggests more).

I wonder if `onClickLexicon` is still needed.


ianb:
Yes, the default behavior when there's a link in the popup is to open the page in a new window (not a new tab in the current window), which is annoying, and that's probably what onClickLexicon was supposed to achieve. activateTab handles that on its own, making onClickLexicon unnecessary.

PR #822, opened 2019-12-27
Save page as [name] intent

### LINKED ISSUE #799: Add "save page as" intent
Maybe just as another match for the current saving.save intent. There's a parameter to `.download()` that we currently leave unset that will force save as.

We could also do "save page as [name]". I'm a little unsure if that's a reasonable way to give a name, given the unreliability of the transcription, but then if we put it in users might just hate it and tell us, which is OK.

maitrella:
Fix #799

### DIFF
--- a/extension/intents/saving/saving.js
+++ b/extension/intents/saving/saving.js
@@ -20,10 +20,11 @@ this.intents.saving = (function() {
       const { html, metadata } = await browser.tabs.sendMessage(activeTab.id, {
         type: "freezeHtml",
       });
-      if (!context.slots.name) {
+      const name = context.slots.name;
+      if (!name) {
         filename = makeFilename(metadata.title, metadata.url, ".html");
       } else {
-        filename = context.slots.name + ".html";
+        filename = name + ".html";
       }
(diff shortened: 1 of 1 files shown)


extension/intents/saving/saving.js
```diff
@@ -18,7 +20,12 @@ this.intents.saving = (function() {
       const { html, metadata } = await browser.tabs.sendMessage(activeTab.id, {
         type: "freezeHtml",
       });
-      const filename = makeFilename(metadata.title, metadata.url, ".html");
+      const name = context.slots.name;
```
ianb: The name should be cleaned with these same regexes that the page title is cleaned with: https://github.com/mozilla/firefox-voice/blob/master/extension/intents/saving/saving.js#L58-L59 (best to extract those into a function).

Unfortunately the download API doesn't automatically clean filenames.

maitrella: Ok. Thanks for the feedback.

ianb:
Alrighty, I'm back, let's land this!

PR #824, opened 2020-01-02
Search, open and read page

### LINKED ISSUE #760: Search/open/read page
Example phrase:

- "read me the wikipedia article on lions"

This should do a Google I'm Lucky search (in a new tab), then start the read action.

This could go in `read.js`

maitrella:
Fix #760

### DIFF
--- a/extension/intents/read/read.js
+++ b/extension/intents/read/read.js
@@ -1,4 +1,4 @@
-/* globals util, content, log, browserUtil */
+/* globals content, log, browserUtil */
 
 this.intents.read = (function() {
   const exports = {};
@@ -55,24 +55,7 @@ this.intents.read = (function() {
         activeTab = await context.createTabGoogleLucky(query);
         await browserUtil.waitForDocumentComplete(activeTab.id);
       }
-      if (!activeTab.url.startsWith("about:reader")) {
-        try {
-          await browser.tabs.toggleReaderMode();
-        } catch (e) {
-          if (
-            e.message &&
-            e.message.includes(
-              "The specified tab cannot be placed into reader mode"
-            )
-          ) {
-            e.displayMessage = "This page cannot be put into Reader Mode";
-          }
-          throw e;
-        }
-        // FIXME: toggleReaderMode just returns immediately so we have to wait to get this to work
-        // Ideally it would give an error or something if it was attached to the wrong kind of tab
-        await util.sleep(5000);
-      }
+      await browserUtil.turnOnReaderMode(activeTab.id);
       await content.lazyInject(activeTab.id, [
         "/intents/read/startNarration.js",
       ]);
--- a/extension/intents/read/read.js
+++ b/extension/intents/read/read.js
@@ -53,7 +53,7 @@ this.intents.read = (function() {
         activeTab = await context.activeTab();
       } else {
         activeTab = await context.createTabGoogleLucky(query);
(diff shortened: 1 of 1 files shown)


maitrella:
@ianb could you tell how to wait for google redirect to be completed?. I think that util.sleep isn't the best way.

ianb:
@maitrella - Is `createTabGoogleLucky()` not working? It had some problems that I thought I fixed... though I guess it may return before the tab has actually loaded (though after the redirect is applied). 

This function could be used (awaited on) to try to force the page to load before you try to toggle it: https://github.com/mozilla/firefox-voice/blob/dd62540ced75101eee1165530e3db0ec921e37fb/extension/browserUtil.js#L122-L127 - but I'm not sure exactly what the state here is before the pause.

maitrella:
Welcome back!. Let me double-check it but I think the function returns before redirect is completed and that's why toggleReaderMode fails. Maybe my computer/connection are really slow.¿?

ianb:
Well, we hope that everything works the same regardless of computer or connection speed, so I wouldn't discount a problem because of that.

My reading of `createTabGoogleLucky` is that it will resolve as soon as the tab object has its new URL, that is not a redirect URL (or `about:blank`). But it will really be _immediately_ after the URL is set on the tab, before any of the page has loaded. I don't know what state is required for `toggleReaderMode` to work, but it's possible that it requires the page to be loaded, or it's possible the page doesn't have to be entirely loaded but needs to be at least a little loaded. `waitForDocumentComplete` waits for the document to be at `document_idle` but `document_end` might be good enough, or heck, maybe even `document_start` (see [here](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/extensionTypes/RunAt)).

maitrella:
Peek 08-01-2020 09-48

You're right. It resolves immediately after the URL is set on the tab.

maitrella:
@ianb Calling toggleReaderMode now works but it also returns immediately (fixme section). I need to increase to 5 seconds in order for intent to work on my computer. ¿Any suggestion?

ianb:
Ah, at first I didn't understand what you mean, but now I do.

So, the problem here is very similar to what is happening in `createTabGoogleLucky`, and the solution will probably be the same. `browserUtil.loadUrl()` is even closer:

```js
  exports.loadUrl = async function loadUrl(tabId, url) {
    await browser.tabs.update(tabId, { url });
    return new Promise((resolve, reject) => {
      function onUpdated(tabId, changeInfo, tab) {
        if (tab.url === url) {
          browser.tabs.onUpdated.removeListener(onUpdated, { tabId });
          resolve(tab);
        }
      }
      browser.tabs.onUpdated.addListener(onUpdated, { tabId });
    });
  };
```

Except you'll need to call toggle instead of browser.tabs.update, and you'll want to check for `tab.url.startsWith("about:reader")`. I'd add another function to browserUtil that toggles reader mode and also waits for it to actually update.

maitrella:
Ok, thanks. I'll wait for that function.

ianb:
@maitrella I've added a function here that should do what you need: ea990248237c01c29587e207ea7165a972b86489

ianb:
@maitrella it looks like this PR has a conflict

maitrella:
@ianb Sorry is my fault I'll try to fix it tomorrow morning.

Issue #830, opened 2020-01-08
Use recorder in popup on Android


ianb:
On desktop we are opening a tab for the microphone and communicating between that tab and the popup. We definitely don't want to do that on Android, and so we want to use the old code path that opens the microphone (getUserMedia).

This is the pref that handles this in the popup:

https://github.com/mozilla/firefox-voice/blob/master/extension/popup/popupController.jsx#L32

buildSettings.js should be updated to turn this pref on for Android. It's possible this codepath has bitrotted since we've used the recorder tab, but it shouldn't be fundamentally broken, just trivially broken.

espertus:
Are you saying that the value of `backgroundTabRecorder` should be different for Android? 

I tried changing it, but, whatever the value (`"1"`, `"0"`, or `false`), I get `Internal error: Error: Incorrect arguments types for tabs.onUpdated.`

ianb:
@espertus Yes, it should be different on Android

There's a problem with `tabs.onUpdated` that was encountered elsewhere as well, where on Android it can't take the `{tabId}` parameter. I fixed one case of it here: https://github.com/mozilla/firefox-voice/commit/6039bce9a31c820d8581c401735c3f0ece214e14

I'm guessing there's more instances of it. E.g., I noticed this function should be updated too (to use `browserUtil.onUpdated{Listen,Remove}` instead of `browser.tabs.onUpdated`): https://github.com/mozilla/firefox-voice/blob/2304eae4ad093b5813c62f4b303440923b535349/extension/browserUtil.js#L25-L36

That's the only other instance I see in the codebase, so I'm guessing loadUrl is the problem here.

PR #838, opened 2020-01-09
Switch to search tab on Android (#830)

espertus:


espertus:
The function `browser.tabs.hide()` is not defined on Android. I guarded one reference to it. A call still remains on line 89 of `search.js` in `_tabHideListener`.

PR #872, opened 2020-01-15
Add intents for handling downloaded items

### LINKED ISSUE #845: Add intents for handling downloaded items
Some things we can do with downloads are noted in https://github.com/mozilla/firefox-voice/issues/807#issuecomment-567152468

We would like to support these followup intents:

- `(open | launch) (download | file)`: calls [browser.downloads.open](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/downloads/open)
- `(show | view | find) (download | file)`: calls [browser.downloads.show](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/downloads/show)

Note we'll have to track the last `downloadId`, which can happen in `saving.js` (which is the only place we're currently creating downloads)

maitrella:
Fix #845

### DIFF
--- a/extension/intents/saving/saving.js
+++ b/extension/intents/saving/saving.js
@@ -32,6 +32,52 @@ this.intents.saving = (function() {
     },
   });
 
+  this.intentRunner.registerIntent({
+    name: "saving.showLastDownload",
+    description: "Shows the last downloaded file in its containing folder",
+    examples: ["Show download"],
+    match: `
+    (show | view | find) (download | file)
+    `,
+    async run(context) {
+      const downloadItems = await browser.downloads.search({
+        limit: 1,
+        orderBy: ["-startTime"],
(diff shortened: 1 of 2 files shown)


maitrella:
@iand I need your feedback on:

1.- I use search() instead of tracking downloadId.
2.- Show intent seems to work fine but open shows "downloads.open may only be called from a user input handler".

ianb:
@maitrella I think your approach of using search is better than remembering the last download – it will work for more cases.

If downloads.open doesn't work without being called from a user input handler (e.g., onclick), then for now we should just not implement it. It's possible that "open" should act the same as "show", since it's the closest we can do to opening the file.

maitrella:
@ianb thanks for the feedback.

PR #874, opened 2020-01-16
Message when pasting into about:newtab

### LINKED ISSUE #853: Give better message when pasting into about:newtab
We can't inject scripts into about:newtab, so if you try to use "paste" on such a tab you'll get an error, as described in #806.

We can't fix the error, but we can catch it and add `exc.displayMessage = "Pasting is not allowed on this page"`.

maitrella:
Fix #853


maitrella:
@ianb thanks for the feedback.

PR #903, opened 2020-01-24
Proof of concept back/forward intents.

### LINKED ISSUE #850: Implement 'back' intent
**Issue Type:** Doesn't behave as expected
**Firefox Version:** Mozilla/5.0 (Macintosh; Intel Mac OS X 10.15; rv:72.0) Gecko/20100101 Firefox/72.0
**Device:** MacBook Pro 15 inch 2016 Catalina
**Location:** CANADA
**GitHub Username**: @jasonrundell
## Steps to Reproduce
1) Speak "Go back"

## Expected Behavior
Expected browser to go back in browser history by 1 step.

## Actual Behavior
Browser goes to https://www.merriam-webster.com/dictionary/back

maitrella:
Trying to fix #850 + npm run format changes.

### DIFF
--- a/extension/intents/navigation/navigation.js
+++ b/extension/intents/navigation/navigation.js
@@ -145,6 +145,34 @@ this.intents.navigation = (function() {
     },
   });
 
+  this.intentRunner.registerIntent({
+    name: "navigation.goBack",
+    description: "Go back to the previous page, if one is available",
+    match: `
+    go back
+    `,
+    async run(context) {
+      const tab = await context.activeTab();
+      await browser.tabs.executeScript(tab.id, {
+        code: "window.history.back();",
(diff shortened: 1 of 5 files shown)


maitrella:
@ianb need feedback on "browser.tabs.executeScript" part.

ianb:
I think using executeScript directly like this should be fine – usually I'd want to use the content.js routines, but in this case the page will immediately unload, and the code is very short, so it seems more reasonable to run it directly.

It looks like some reformatting got caught up in this. We should probably fix those, but keep it separate from this commit, so I'd encourage only adding changes you've added.

maitrella:
@ianb Thanks for the feedback!.

PR #936, opened 2020-02-04
Format changes after "npm run format"

maitrella:


ianb:
Cool, thanks for doing this chore!

PR #965, opened 2020-02-06
Fix missing host permission working with PDF pages.

### LINKED ISSUE #954: Scrolling (etc) doesn't work on a PDF page
For instance, load [this](https://www.cs.rochester.edu/research/cisd/pubs/2007/allen-et-al-plow-aaai2007.pdf) and try to scroll down. You'll get a missing host permission.

I think if we add a permission for `"resource://pdf.js/"` to manifest.json.ejs that it will fix this.

maitrella:
Fix #954 host permission error with pdf.js

### DIFF
--- a/extension/manifest.json.ejs
+++ b/extension/manifest.json.ejs
@@ -101,7 +101,8 @@
     "downloads",
     "browserSettings",
     "search",
-    "tabHide"
+    "tabHide",
+    "resource://pdf.js/"
   ],
   "browser_action": {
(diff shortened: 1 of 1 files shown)


ianb:
Seems right, did it work when you tried it?

maitrella:
@ianb it doesn't work.

ianb:
Well, I think the host permission change is fine, it just turns out there's more to this bug, so I'll land this.

Issue #914, opened 2020-01-24
Bookmark page


awallin:
Based on telemetry users are trying to bookmark a page using their voice.

"save this to bookmarks"
"bookmark this"
"bookmark this page"
"add this page to bookmarks"

ianb:
This could be done with the [bookmarks.create()](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/bookmarks/create)

I think simpler is better, so we can use these phrases to create a simple bookmark placed in no specific location. The `pageMetadata.js` handler can get a good title and URL.

paulZhongym:
@ianb Hi, are there any other features needed in this?

ianb:
I think this is fixed with #967. It would be possible to add some new phrases that set the bookmarks title and such, but I'm not sure that kind of input is accurate enough to work well.

PR #967, opened 2020-02-06
Bookmark page intent

maitrella:
Creates a bookmark for this page, placing it in the default folder.


maitrella:
Minimal version for #914

Issue #996, opened 2020-02-14
Use window.docShell.doCommand() for more


ianb:
We can do things ([like in this file](https://dxr.mozilla.org/mozilla-central/source/editor/libeditor/tests/test_selection_move_commands.html)) like `cmd_scrollPageDown`, and this is going to act exactly like hitting the Page Down button (which is different than `window.scrollBy()`).

It should enable a bunch of simple commands with very little code.

[This search](https://dxr.mozilla.org/mozilla-central/search?q=specialpowers&redirect=false) shows a bunch of tests that use these, and can be examples of the available commands.

Note this will have to go in `api.js` since it's privileged code. We shouldn't pass a string unchecked from the extension to `doCommand()`, or we should whitelist those commands.

ianb:
I've had fairly bad luck using `window.goDoCommand`, but clicking `<command>` elements does work for a couple things (like #1142)

PR #1003, opened 2020-02-14
Try out #996, using doCommand('cmd_scrollPageDown') etc

ianb:
Note this only works once for as-yet-unknown reasons


ianb:
While page down hasn't been working, this is otherwise working in master.

Issue #1004, opened 2020-02-17
After click on mic icon "Doorhanger" opens and closes immediately


maitrella:
STR:
    Click on the mic icon from the browser toolbar.

Expected result:
"Doorhanger" opens

Actual result:
"Doorhanger" opens and closes immediately 

Notes: Reproduced on Ubuntu 19.10 with 74.0a1 (2020-02-09) (64-bit)

![Peek 17-02-2020 11-22](https://user-images.githubusercontent.com/5709917/74646005-4b091300-5179-11ea-99e2-f541b096a924.gif)

maitrella:
Same happens with 75.0a1 (2020-02-14) (64-bit)

alexandra-martin:
I am experiencing a similar issue, if I install the add-on from the "Install dev version" link. If I click normally on the mic icon nothing happens, but on Firefox 73.0 if I click very fast several times I see a little bit of the doorhanger.
Tried it on Windows 10x64 with Firefox 73.0 and 75.0a1 (2020-02-17).
The add-on installed from the "Install production version" link, works fine, but don't know if it's the latest version.

![2](https://user-images.githubusercontent.com/54401091/74737154-f343d800-525c-11ea-8628-82a0daff3ce5.gif)

maitrella:
The problem appears with the latest code in master branch.

ianb:
@mairella @alexandra-martin I'm not able to reproduce it here. Is there any error message in the Browser Console?

maitrella:
I'll check tomorrow morning. To test a new branch (or master) before npm install, npm run format and npm start I erase node_modules and Firefox profile directories. Doing this I can always reproduce the issue with code in master branch.

maitrella:
@ianb don't take into account my comment  I wrote it before your PR.

PR #1005, opened 2020-02-17
Use SpaCy stopwords

### LINKED ISSUE #999: Use SpaCy stopwords
In `extension/background/language/english.js` we have a list of stopwords. SpaCy [has a bigger list](https://github.com/explosion/spaCy/blob/e0cf4796a505bd26e8fcb95d23b89fd7eca3be0a/spacy/lang/en/stop_words.py) and maybe we should use that.

maitrella:
Fix #999

### DIFF
--- a/extension/background/language/english.js
+++ b/extension/background/language/english.js
@@ -51,17 +51,46 @@ upward up ward
 }
 
 for (let line of `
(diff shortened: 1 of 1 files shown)


maitrella:
@ianb #1004 doesn't allow me to test the stopwords change.

PR #1094, opened 2020-02-26
Added more tab examples

awallin:


extension/views/lexicon.html.ejs
```diff
@@ -207,14 +207,17 @@
 
         <div class="example">
           <div class="intent">
-            Find tab
+            Find & switch tabs
```
ianb: Can you change this to `&amp;` ?

awallin: Done

PR #1129, opened 2020-02-29
Fix #1081, switch to next/previous window 

### LINKED ISSUE #1081: Next/previous window intents
In #1078 we got tab switching intents, we could have the same kind of intents for windows.

For example, "switch to next window". This would use the same phrases as tabs, but with "window"

yuanyu90221:


extension/intents/window/window.js
```diff
@@ -0,0 +1,42 @@
...
+
+function findTargetWindowId(windowArray, currentWindowId, direction) {
+    const len = windowArray.length;
+    // find currentWindowId postion in array
+    const currentWindowIndex = windowArray.findIndex((window) => (window.id === currentWindowId));
+    let targetIndex = 0;
+    if ( direction === "next" ) {
+      targetIndex = Math.floor((currentWindowIndex + 1) % len);
+    } else {
+      targetIndex = Math.floor((currentWindowIndex - 1 + len) % len);
```
ianb: `Math.floor` won't be necessary here, everything is an integer so it should stay an integer.

extension/intents/window/window.js
```diff
@@ -0,0 +1,42 @@
+import * as intentRunner from "../../background/intentRunner.js";
+
+function findTargetWindowId(windowArray, currentWindowId, direction) {
+    const len = windowArray.length;
+    // find currentWindowId postion in array
+    const currentWindowIndex = windowArray.findIndex((window) => (window.id === currentWindowId));
+    let targetIndex = 0;
```
ianb: This doesn't need to be initialized to anything, since it will definitely be reset below.

extension/intents/window/window.js
```diff
@@ -0,0 +1,42 @@
...
+      targetIndex = Math.floor((currentWindowIndex + 1) % len);
+    } else {
+      targetIndex = Math.floor((currentWindowIndex - 1 + len) % len);
+    }
+    return windowArray[targetIndex].id;
+  }
+// error handle for
+function onError(error) {
+  // console.log(`Error: ${error}`);
+}
```
ianb: There doesn't need to be any error handler, if there's an error in `.run()` the intentRunner will catch it.

extension/intents/window/window.js
```diff
@@ -0,0 +1,42 @@
...
+    name: "window.switch",
+    async run(context) {
+        // get current activeTab.windowId
+        const activeTab = await context.activeTab();
+        const currentWindowId = activeTab.windowId;
+        // get direction parameter
+        let direction = "next";
+        if ( context.parameters ) {
+           direction = context.parameters.direction;
+        }
```
ianb: This will definitely be set, so you can just do `let direction = context.parameters.direction` without the if.

extension/intents/window/window.js
```diff
@@ -0,0 +1,42 @@
...
+        try {
+          // getAll normal window
+          const gettingAll = await browser.windows.getAll({windowTypes: ["normal"]});
+          // find target windowId
+          const targetWindowId = findTargetWindowId(gettingAll, currentWindowId, direction);
+          // set target window focuse true
+          await browser.windows.update(targetWindowId, {focused: true});
+        } catch (err) {
+          onError(err);
+        }
```
ianb: You can remove the entire try/catch, and the comments all match the code so they aren't really necessary either.

extension/intents/window/window.toml
```diff
@@ -0,0 +1,6 @@
+[window.switch]
+description = "Switch the current window"
+match = """
+  switch to next window [direction=next]
+  switch to previous window [direction=back]
```
ianb: I'd do the phrase something like:

```
match = """
(switch | activate | focus) (the | to |) next window [direction=next]
(switch | activate | focus) (the | to |) (previous | last) window [direction=back]
"""

ianb:
Looking good, thank you for your contribution!

Issue #1113, opened 2020-02-27
Intent: paste to notes


ianb:
This would go with the other note-taking intents, and would just paste whatever happened to be in the clipboard. Mostly it would just call `document.execCommand("paste")` once the field was focused.

ianb:
For this you'll want to look in `notes.js` and create a new intent, very similar to the other intents.

alexandra-martin:
Verified on Mac 10.14 and Win10x64 with Firefox Nightly 76.0a1 (2020-03-09).

![8](https://user-images.githubusercontent.com/54401091/76313540-cfb10200-62dd-11ea-9666-d3232e9b7fe9.gif)

PR #1130, opened 2020-02-29
paste to notes intent

Manasa2850:
Solves issue #1113 
@ianb please let me know if my approach is correct.

### DIFF
--- a/extension/intents/notes/notes.js
+++ b/extension/intents/notes/notes.js
@@ -90,8 +90,7 @@ intentRunner.registerIntent({
   async run(context) {
     await checkHasTab();
     const success = await browser.tabs.sendMessage(writingTabId, {
-      type: "addText",
-      text: document.execCommand("paste"),
+      type: "pasteText",
     });
     if (!success) {
       const e = new Error("Could not paste text");
--- a/extension/intents/notes/notes.js
+++ b/extension/intents/notes/notes.js
@@ -84,3 +84,19 @@ intentRunner.registerIntent({
     await context.makeTabActive(writingTabId);
   },
 });
+
+intentRunner.registerIntent({
+  name: "notes.paste",
+  async run(context) {
+    await checkHasTab();
+    const success = await browser.tabs.sendMessage(writingTabId, {
+      type: "addText",
+      text: document.execCommand("paste"),
+    });
+    if (!success) {
+      const e = new Error("Could not paste text");
+      e.displayMessage = "Could not paste text";
+      throw e;
+    }
+  },
+});
--- a/extension/intents/notes/notes.toml
+++ b/extension/intents/notes/notes.toml
@@ -19,7 +19,7 @@ phrase = "Make note of this page"
 [notes.paste]
 description = "Focus the tab where you want to paste"
 match = """
-  (paste) (on |) (this |) note (for me |)
+  paste (on | in | into |) note{s} (for me |)
 """
 
 [[notes.paste.example]]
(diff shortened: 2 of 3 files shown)


extension/intents/notes/notes.js
```diff
@@ -84,3 +84,19 @@ intentRunner.registerIntent({
...
   },
 });
+
+intentRunner.registerIntent({
+  name: "notes.paste",
+  async run(context) {
+    await checkHasTab();
+    const success = await browser.tabs.sendMessage(writingTabId, {
+      type: "addText",
+      text: document.execCommand("paste"),
```
ianb: `document.execCommand("paste")` doesn't return anything useful (it may return true/false for success/failure, but even that seems unreliable). Instead it actually performs the paste.

So to get this to paste in a document you'll need to add to `extension/intents/notes/contentScript.js` – a new `communicate.register("paste", () => {...})`

That command should call `focus()` and then `document.execCommand("paste")`. It could return the value of document.execCommand, but it's very possible that's going to give false errors to the user.

extension/intents/notes/notes.toml
```diff
@@ -16,6 +16,15 @@ match = """
 [[notes.addLink.example]]
 phrase = "Make note of this page"
 
+[notes.paste]
+description = "Focus the tab where you want to paste"
+match = """
+  (paste) (on |) (this |) note (for me |)
```
ianb: I think a better match phrase might be:

```
match = """
  paste (on | in | into |) note{s} (for me |)
"""

File: extension/intents/notes/notes.js (hunk 1 above)
Comment (Manasa2850) [reply extension/intents/notes/notes.js]: Thanks! I'll make the changes

Comment (Manasa2850) [reply extension/intents/notes/notes.js]: communicate.register("paste", () => {...}) is similar to that of addText right?

Comment (ianb) [reply extension/intents/notes/notes.js]: Yep!

File: extension/intents/notes/notes.toml [hunk 3]
```diff
@@ -16,6 +16,15 @@ match = """
...
 phrase = "Make note of this page"
 
+[notes.paste]
+description = "Focus the tab where you want to paste"
+match = """
+  paste (on | in | into |) note{s} (for me |)
+"""
+
+[[notes.paste.example]]
+phrase = "Paste everything present on the clipboard to notes"
```
ianb: The example here should be something that is matched by the phrase above. So `Paste into notes` would be a good example.

`npm run jest` will also test examples, and should be giving you an error with this code (since the example isn't actually matched by the match phrase).

extension/intents/notes/notes.toml
```diff
@@ -16,6 +16,15 @@ match = """
 [[notes.addLink.example]]
 phrase = "Make note of this page"
 
+[notes.paste]
+description = "Focus the tab where you want to paste"
```
ianb: This won't focus the tab, a better description would just be "Pastes the clipboard into your note"

Manasa2850: Yup I've made the changes.
Is there a naming convention for the descriptions and phrases? Wanted to learn about that so I don't repeat these mistakes again.

ianb: The description isn't particularly important, it's just there for developers. But it's always good to be accurate ;)

Examples should usually be easy for the user to understand – so even if you match a short form, you might put in a longer form of the phrase in the example so that it's obvious what it will do. There's no extra information given with examples, so they have to be self-explanatory.

For phrase matching it's just a matter of trying to think up different ways someone might phrase the command. It's more art than science at this point!

Manasa2850: Thanks @ianb 
I'll keep this in mind while working :)

PR #1137, opened 2020-03-02
 Adds remove action to bookmark intent file

### LINKED ISSUE #1132: Bookmark intent sheet does not contain step to remove bookmark
In addition to creation and opening of bookmark. Removing bookmark from current page will be good add-on feature.

GangaChatrvedi:
Fixes #1132

 - [x] Example utterance are checked and tested.

### DIFF
--- a/extension/intents/bookmarks/bookmarks.js
+++ b/extension/intents/bookmarks/bookmarks.js
@@ -72,3 +72,15 @@ intentRunner.registerIntent({
     }
   },
 });
+
+intentRunner.registerIntent({
+  name: "bookmarks.remove",
+  async run(context) {
+    const activeTab = await context.activeTab();
+    const metadata = await pageMetadata.getMetadata(activeTab.id);
+    const bookmarks = await browser.bookmarks.search({ url: metadata.url });
+    for (const bookmark of bookmarks) {
+      await browser.bookmarks.remove(bookmark.id);
+    }
+  },
+});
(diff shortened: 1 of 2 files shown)


extension/intents/bookmarks/bookmarks.js
```diff
@@ -72,3 +72,15 @@ intentRunner.registerIntent({
     }
   },
 });
+
+intentRunner.registerIntent({
+  name: "bookmarks.remove",
+  async run(context) {
+    const activeTab = await context.activeTab();
+    const metadata = await pageMetadata.getMetadata(activeTab.id);
+    const bookmarks = await browser.bookmarks.search({ url: metadata.url });
```
ianb: This part is a little subtle: pageMetadata gets the [canonical URL](https://yoast.com/rel-canonical/), while `activeTab.url` will be the URL in the URL bar. For many pages this will be the same, but it would be better to use `activeTab.url` here since it's the most likely thing that was bookmarked.

extension/intents/bookmarks/bookmarks.js
```diff
@@ -72,3 +72,15 @@ intentRunner.registerIntent({
...
+    for (const bookmark of bookmarks) {
+      await browser.bookmarks.remove(bookmark.id);
```
ianb: This should give an error message if `bookmarks` is empty (look for `displayMessage` for examples).

extension/intents/bookmarks/bookmarks.toml
```diff
@@ -41,3 +41,18 @@ expectSlots = { query = "query" }
 phrase = "open bookmark query for me in tab"
 test = true
 expectSlots = { query = "query" }
+
+[bookmarks.remove]
+description = "Remove bookmark from the default folder (Other Bookmarks)"
+match = """
+  remove bookmark (from |) (this |) (page | site | tab |) (for me |)
+  (remove | delete) (this |) (page | site | tab |) from bookmarks
```
ianb: You could use `(remove | delete)` in the first phrase as well

ianb:
Looks good, thanks!

Issue #1038, opened 2020-02-20
[wakeword] Update Porcupine library


ianb:
I believe the npm package is out of date, and instead we should be installing it from source

ianb:
This really went sideways, not sure what exactly to do about it. Fails in several ways.

ianb:
We decided to remove Porcupine (#1596), so this is moot

PR #1141, opened 2020-03-02
WIP #1038, updates .ppn and support files

ianb:
This copies the files in instead of installing via npm, as this appears to be the preferred technique now

Does not actually work


ianb:
I don't think we're going to move forward on this

PR #1150, opened 2020-03-03
Copy image indent

### LINKED ISSUE #1007: "Copy image" intent
This intent would find the "best" image on the page, and copy it. Probably the image that is visible in the viewport (50%) that has the largest area.

JanviMahajan14:
Fixes #1007

### DIFF
--- a/extension/intents/clipboard/clipboard.js
+++ b/extension/intents/clipboard/clipboard.js
@@ -72,9 +72,9 @@ intentRunner.registerIntent({
 });
 
 intentRunner.registerIntent({
-  name: "clipboard.copyImageLink",
+  name: "clipboard.copyImage",
   async run(context) {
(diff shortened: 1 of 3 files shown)


extension/intents/clipboard/clipboard.js
```diff
@@ -71,6 +71,14 @@ intentRunner.registerIntent({
   },
 });
 
+intentRunner.registerIntent({
+  name: "clipboard.copyImageLink",
```
ianb: One other thing – this (I believe) is copying the image to the clipboard, not a link to the image. So the intent should be named accordingly, `"clipboard.copyImage"`

JanviMahajan14:
@ianb i have updated the changes suggested by you. Hope you like it :)

ianb:
Look good!

PR #1151, opened 2020-03-03
Add a script that will enumerate all phrases by using the patterns

ianb:
Also adds support for importing JavaScript modules from Node using babel-node

### DIFF
--- a/bin/enumerate-phrases.js
+++ b/bin/enumerate-phrases.js
@@ -1,0 +1,62 @@
+/* eslint-disable no-console */
+const { metadata } = require("../extension/intents/metadata.js");
+const { entityTypes } = require("../extension/background/entityTypes.js");
+const {
+  compile,
+  splitPhraseLines,
+} = require("../extension/background/language/compiler.js");
+
+const args = process.argv.slice(2);
+
+function includeIntent(category, subname) {
+  if (!args.length) {
+    return true;
+  }
+  const s = `${category}.${subname}`;
+  for (const arg of args) {
+    if (s.startsWith(arg) || s.endsWith(arg)) {
+      return true;
+    }
+  }
+  return false;
+}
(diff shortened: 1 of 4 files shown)


ianb:
@reuben I ended up replacing untyped slots with their name (e.g., `QUERY`). There's a lot of slots, I don't know what the best thing to do there is. We could leave out any words at all? Include some sample queries?

It's about 74k phrases, 580k words. It does not include any stopwords unless they were specifically included in the phrase matcher.

reuben:
@ianb amazing! I'm on vacation right now and about to travel to a place without much reception, but I'll take a closer look when I'm back next week. I think the untyped slots are very important to get right as that's where LMs built from general text corpora might fail, due to not giving appropriate weight to e.g. popular site names. There's a few simple substitutions that can be done for select slots, for example using the top N most visited website names for tab and site names. And for the others I'm thinking of good ways to leverage the product data we have, but nothing concrete yet. As long as the untyped slots can be easily recognized in the output for post processing, these things can be done as additional logic on top of the enumerate script here.

In any case, this is an excellent start, the first thing I want to try is simply add this language  to a general text corpus and see how the resulting LM behaves.

ianb:
I think it might make sense to simply split on the untyped slots, training the fragment before and after the slot. Then if we could get a good set of search terms, those could be trained separately. Unless you think that having a holistic sense of the entire sentence is important.

We're trying to get usable actual utterances from Firefox Voice as well, but that's still an ongoing task. But yeah, it's possible other products have already produced usable lists of queries we can train against.

(I guess arguably while most of the untyped slots are search queries, they are different kinds of search depending on the phrase: e.g., `play [query]` will be an artist of song, while `go to [query]` will be some short description of a site, and `search [query]` will be a general search.)

Issue #1097, opened 2020-02-26
[New intent] Downloads menu


awallin:
Users are requesting an action to open the downloads menu.
- "Show downloads"
- "Open downloads menu"
- "Downloads menu"

PR #1155, opened 2020-03-04
Open downloads window intent via experiments.

maitrella:
Proof of concept of #1097 it opens downloads window but no download menu. 
@ianb I need feedback.


ianb:
Yeah, I haven't figured out how to open menus yet, Firefox does that in some different fashion. I believe this is the equivalent of **Tools > Downloads**, which is fine. But we'll want to understand the other menus eventually, especially "Send Tab to Device"

@awallin do you have any phrase suggestions for this?

awallin:
I documented a few download menu utterance I came across in telemetry in #1097

Issue #919, opened 2020-01-24
Close Firefox utterance


awallin:
User wanted to shut down Firefox by saying "close Firefox" or "close the browser"

awallin:
Seeing this same request multiple times in the post action telemetry feedback.

maitrella:
@ianb [window.close](https://developer.mozilla.org/en-US/docs/Web/API/Window/close) "Scripts may not close windows that were not opened by script".  Maybe is impossible to implement a close browser intent.

ianb:
I think this is intended to close the entire program, so there's no easy API to do it.

maitrella:
In my mind: "If you close all the open windows them firefox closes too" XDDDD

ianb:
True, on Windows it probably does – on Mac it just leaves the app open with no windows.

awallin:
Another feature request came in asking for "exit Firefox" / "quit firefox" / "shutdown firefox"

https://github.com/mozilla/firefox_voice_community/issues/292

GangaChatrvedi:
@ianb @awallin as pr is merged for this. should we close this issue?

ianb:
@GangaChatrvedi Good catch, yes

alexandra-martin:
Verified on Mac 10.14 with Firefox Nightly 76.0a1 (2020-04-01) for the current intent.
#1319 is a follow up for possibly having the confirmation window before Firefox is closed.
#1457 is a follow up for adding "(the) browser" in the intent.

![1](https://user-images.githubusercontent.com/54401091/78245359-039fd180-74f0-11ea-960a-bfe50a6b907d.gif)

PR #1156, opened 2020-03-04
Close Firefox intent via experiments.

maitrella:
Proof of concept #919 
@ianb I need feedback.

Tested only on linux sorry.


ianb:
@maitrella Oops, it looks like this PR conflicted with your last PR

maitrella:
I'll try to fix it next monday.  Thanks for notify me.

PR #1163, opened 2020-03-06
Format linting test

### LINKED ISSUE #1144: Make format linting part of tests
With more PRs coming in, it might be nice to make these part of the tests.

`prettier --check` doesn't actually return with a non-zero code. There's [prettylint](https://github.com/ikatyang/prettylint).

One nuisance is that anything that's not in git shouldn't be linted. But .prettierignore has a couple files in source control that shouldn't be linted either. Maybe we should just add `// prettier-ignore` to those two files (voiceSchema.js and catcherAsyncSetup.js).

dave-ok:
Fixes #1144 
- [x] Install prettylint

- [x] Add script for pretty lint specifying ignore path

- [x] Add //pretty-ignore to voiceSchema.js and catcherAsyncSetup.js


ianb:
The [test is failing](https://circleci.com/gh/mozilla/firefox-voice/2384?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link), it looks like `.gitignore` isn't being used by prettylint

dave-ok:
Ok will get on it right away

ianb:
Both this and #1249 fix the issue. I think adding it to eslint is probably better, but I'll merge/revert this one and then add that one since they are both valid fixes.

dave-ok:
Thank you

Issue #1083, opened 2020-02-25
Submit form intent


ianb:
The submit form intent might simulate an Enter key, or could try to find the most plausible form to submit and call `.submit()` (though Enter would be better).

Should match at least `submit (form |)`, but maybe there's other phrases too.

This would be a new intent in `forms.js`

Manasa2850:
I want to work on this

ianb:
You got it!

Manasa2850:
@ianb how do I reproduce this in my browser?
I'm not quite sure how to view the form intent in my browser

ianb:
I'd recommend finding a few forms on the web and testing them. The search form on github.com is one:

<img width="592" alt="image" src="https://user-images.githubusercontent.com/44390/75892526-1642c000-5df7-11ea-928c-42b824174284.png">

This comment form might also be a good example, as Enter _won't_ work here.

You'll probably want to look at `document.activeElement` and then identify the form based on that (probably `document.activeElement.form` though you should probably fall back to looking through parentNodes, as not every active element will have that `.form` attribute)

Manasa2850:
Ok. I'll try this out. Thanks!

PR #1164, opened 2020-03-06
submit form intent

Manasa2850:
Fixes issue #1083 
@ianb I'm a little doubtful whether `e.submitTextInput();` is the right approach.
Please let me know if that's correct. Thanks!

### DIFF
--- a/extension/intents/forms/forms.toml
+++ b/extension/intents/forms/forms.toml
@@ -29,7 +29,7 @@ phrase = "Focus next field"
 [forms.formSubmit]
 description = "Submit the form field"
 match = """
-  focus submit (form |)
+  submit (form |)
 """
 
 [[forms.formSubmit.example]]
--- a/extension/intents/forms/forms.toml
+++ b/extension/intents/forms/forms.toml
@@ -26,6 +26,15 @@ match = """
 [[forms.focusNext.example]]
 phrase = "Focus next field"
 
+[forms.formSubmit]
+description = "Submit the form field"
+match = """
+  focus submit (form |)
+"""
+
+[[forms.formSubmit.example]]
+phrase = "Submit Form"
+
 [forms.focusPrevious]
 description = "Focus the previous field"
 match = """
--- a/extension/intents/forms/formsContentScript.js
+++ b/extension/intents/forms/formsContentScript.js
@@ -176,6 +176,14 @@ this.dictationContentScript = (function() {
     return focusDirection(1);
   });
 
+  communicate.register("formSubmit", async message => {
+    onInputKeyPress = event(e) => {
+      if (event.key === "Enter") {
+        e.submitTextInput();
+      }
+    };
+  });
+
   communicate.register("focusPrevious", async message => {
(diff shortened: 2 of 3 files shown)


extension/intents/forms/forms.toml
```diff
@@ -26,6 +26,15 @@ match = """
 [[forms.focusNext.example]]
 phrase = "Focus next field"
 
+[forms.formSubmit]
+description = "Submit the form field"
+match = """
+  focus submit (form |)
```
ianb: The word "focus" shouldn't be here. `Submit (form |)` should be enough.

Probably the local tests are failing right now in this case too (unfortunately I haven't figured out how to run these particular tests in CircleCI). Specifically the example you give below won't be matched by this pattern. You can test this by running `npm run jest`

extension/intents/forms/formsContentScript.js
```diff
@@ -176,6 +176,14 @@ this.dictationContentScript = (function() {
     return focusDirection(1);
   });
 
+  communicate.register("formSubmit", async message => {
+    onInputKeyPress = event(e) => {
+      if (event.key === "Enter") {
+        e.submitTextInput();
+      }
+    };
```
ianb: I don't think this will do anything? Or at least I'm unclear what it will do, it seems to be creating a global function, but I don't think anything will ever call this function.

What should happen here is that it actually submits the form. Probably more like `document.activeElement.form.submit()` – though there might be some caveats depending on what element is active.

extension/intents/forms/forms.toml
Manasa2850: So what shall I change the example to so that we don't get CircleCI errors?

ianb: No... if you change the match pattern and remove "focus" it should work.

Manasa2850:
@ianb I felt this approach might be better since every page need not necessarily have an enter/submit button

extension/intents/forms/formsContentScript.js
```diff
@@ -176,6 +176,11 @@ this.dictationContentScript = (function() {
     return focusDirection(1);
   });
 
+  communicate.register("formSubmit", async message => {
+    document.activeElement.form.submit();
```
ianb: This is looking pretty good, though it should also check and give an error here: sometimes `document.activeElement.form` will be null. Not infrequently actually, anytime the active element isn't a form element this will be true. So for instance if you click on the background of the page and then try to submit a form you'll get this:

image

If you look for `displayMessage` you'll see the pattern to throw a special exception that will give the user an error message. In this case the user should see something like "Please focus on an input field before trying to submit"

Manasa2850:
@ianb I've used `null` itself and not `NULL` in my latest commit. I checked it out and it seems to be working fine.

Manasa2850:
I've formatted the code!

ianb:
Great, looking good now!

Manasa2850:
Thanks for the help @ianb !!

PR #1165, opened 2020-03-06
Open and go new tab/new window

### LINKED ISSUE #761: Open and go new tab/new window
Some examples of utterances we could support:

- "open a new window and go to the Washington Post"
- "open a new window and go to the New York Times"
- "open a new tab to the New York Times"
- "open the New York Times in a new tab/window"

I think these should go in the `navigation.navigate` intent, and add `[where=newtab]` or `[where=newwindow]` as parameters, and adjust the intent as appropriate to make use of those parameters.

JanviMahajan14:
Fixes #761

### DIFF
--- a/extension/intents/navigation/navigation.js
+++ b/extension/intents/navigation/navigation.js
@@ -9,7 +9,7 @@ import * as browserUtil from "../../browserUtil.js";
 const QUERY_DATABASE_EXPIRATION = 1000 * 60 * 60 * 24 * 30; // 30 days
 const queryDatabase = new Map();
 
-function database(query, tab, url) {
+function saveTabQueryToDatabase(query, tab, url) {
   queryDatabase.set(query.toLowerCase(), {
     url,
     date: Date.now(),
@@ -22,9 +22,10 @@ function database(query, tab, url) {
         url: newTab.url,
         date: Date.now(),
       });
+      saveQueryDatabase();
     }
-    saveQueryDatabase();
   }, 1000);
+  saveQueryDatabase();
 }
 
 intentRunner.registerIntent({
@@ -34,21 +35,21 @@ intentRunner.registerIntent({
     const where = context.slots.where;
     const cached = queryDatabase.get(query.toLowerCase());
     if (where === "window") {
-       if (cached) {
-        await browser.windows.create({url: cached.url});
+      if (cached) {
+        await browser.windows.create({ url: cached.url });
       } else {
         await browser.windows.create({});
         const tab = await context.createTabGoogleLucky(query);
         const url = tab.url;
-        database(query, tab, url);
+        saveTabQueryToDatabase(query, tab, url);
       }
     } else if (cached) {
-        await context.openOrFocusTab(cached.url);
-      } else {
-          const tab = await context.createTabGoogleLucky(query);
-          const url = tab.url;
-          database(query, tab, url);
-        }
+      await context.openOrFocusTab(cached.url);
+    } else {
+      const tab = await context.createTabGoogleLucky(query);
+      const url = tab.url;
+      saveTabQueryToDatabase(query, tab, url);
+    }
     context.done();
   },
 });
--- a/extension/intents/navigation/navigation.js
+++ b/extension/intents/navigation/navigation.js
@@ -13,28 +13,36 @@ intentRunner.registerIntent({
   name: "navigation.navigate",
   async run(context) {
     const query = context.slots.query;
+    const where = context.slots.where;
     const cached = queryDatabase.get(query.toLowerCase());
-    if (cached) {
-      await context.openOrFocusTab(cached.url);
-    } else {
-      const tab = await context.createTabGoogleLucky(query);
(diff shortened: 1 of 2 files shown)


extension/intents/navigation/navigation.toml
```diff
@@ -63,6 +63,7 @@ match = """
 description = "Navigate directly to a site, using Google's I'm Feeling Luck and the query"
 match = """
   (bring me | take me | go | navigate | show me | open) (to | find |) (page |) [query]
+  (bring me | take me | go | navigate | show me | open) (to | find |) (page |) (in |) (a |) (new |) [where] (and go | and) (to | find | open) [query]
```
ianb: `[where]` here will catch anything, so it could also catch "open a new bucket to the new york times"

Adding a new type to `entityTypes.js` would fix this, though that would also be fine in a followup.

extension/intents/navigation/navigation.js
```diff
@@ -13,28 +13,36 @@ intentRunner.registerIntent({
...
+          // Sometimes there's a very quick redirect
+          setTimeout(async () => {
+            const newTab = await browser.tabs.get(tab.id);
+            if (newTab.url !== url) {
+              queryDatabase.set(query.toLowerCase(), {
+                url: newTab.url,
+                date: Date.now(),
+              });
+            }
+          }, 1000);
```
ianb: All of this code should really be run if `where` is a tab or window. So really while there's that `browser.windows.create({})` code that's conditional on `where`, the rest can and should be run regardless.

extension/intents/navigation/navigation.js
```diff
@@ -13,28 +13,36 @@ intentRunner.registerIntent({
...
+          saveQueryDatabase();
```
ianb: This isn't your bug, but a bug of mine I noticed while reading the code: `saveQueryDatabase()` should be run inside the setTimeout if `queryDatabase` is modified.

extension/intents/navigation/navigation.js
JanviMahajan14: Actually this is creating problem as **tab.id** and **url** will not be defined if put outside the else as the scope is limited within else block. Another way to resolve this is to add a setTimeOut fn within the where===window conditional also . But it is rewriting of code . What you do this I should do?

extension/intents/navigation/navigation.js
JanviMahajan14: i will fix this

JanviMahajan14:
@ianb i have moved the code required for both the cases to a separate function . Do let me know if this is correct?

extension/intents/navigation/navigation.js
```diff
@@ -9,32 +9,46 @@ import * as browserUtil from "../../browserUtil.js";
 const QUERY_DATABASE_EXPIRATION = 1000 * 60 * 60 * 24 * 30; // 30 days
 const queryDatabase = new Map();
 
+function database(query, tab, url) {
```
ianb: This name isn't very descriptive, maybe it could be `saveTabQueryToDatabase(...)` ?

extension/intents/navigation/navigation.js
```diff
@@ -9,32 +9,46 @@ import * as browserUtil from "../../browserUtil.js";
...
+    const newTab = await browser.tabs.get(tab.id);
+    if (newTab.url !== url) {
+      queryDatabase.set(query.toLowerCase(), {
+        url: newTab.url,
+        date: Date.now(),
+      });
+    }
+    saveQueryDatabase();
+  }, 1000);
+}
```
ianb: There should also be a call to `saveQueryDatabase()` at the end of this function.

extension/intents/navigation/navigation.js
```diff
@@ -9,32 +9,46 @@ import * as browserUtil from "../../browserUtil.js";
...
+  // Sometimes there's a very quick redirect
+  setTimeout(async () => {
+    const newTab = await browser.tabs.get(tab.id);
+    if (newTab.url !== url) {
+      queryDatabase.set(query.toLowerCase(), {
+        url: newTab.url,
+        date: Date.now(),
+      });
+    }
+    saveQueryDatabase();
```
ianb: This call to `saveQueryDatabase()` should be inside the if block because it only needs to be called if we actually update the database.

extension/intents/navigation/navigation.js
JanviMahajan14: Sure @ianb I will update this

extension/intents/navigation/navigation.js
JanviMahajan14: Even I like this name too I will update this

JanviMahajan14:
@ianb i have done the required changes . Have a look.

extension/intents/navigation/navigation.js
ianb: It actually occurred to me later that your code was actually correct, but this code is also correct, so no further updates needed ;)

ianb:
Thanks, look great!

PR #1171, opened 2020-03-07
fixes #422 - fixed stop, pause, skip commands

### LINKED ISSUE #422: "Pause", "Stop", "Next", "Skip" commands not working for videos if Youtube tab was firstly opened in a channel page
### Prerequisites:
Mic and autoplay permissions are enabled.
Make sure that when making the Youtube commands, the Youtube tab was first opened in a channel page.

### STR:
1. Use the shortcut or click on the mic icon from the browser toolbar and make a "Play" command, e.g. "Play Foo on Youtube".
2. On the opened Youtube channel page tab, select one or more videos from the present page.
3. Use the shortcut or click on the mic icon from the browser toolbar in the Youtube tab.
4. Say or write a "Pause" command.

### Expected result:
The Youtube video is paused.

### Actual result:
The Youtube video continues playing.

### Notes:
Reproduced on Mac 10.14.6 and Win10x64 with Firefox Nightly 71.0a1 (64-bit).
Reproduced also for the "Stop", "Skip" and "Next" commands. (second gif)
It seems the commands will not work in this kind of tab, even if several videos are opened.

melvin2016:
Fixes #422 - Added the fix for Stop, Pause Commands not working. Thanks to @ianb  for pointing out the start of the issue.
Screen-Recording-2020-03-07-at-7

### DIFF
--- a/extension/intents/music/music.toml
+++ b/extension/intents/music/music.toml
@@ -68,7 +68,7 @@ phrase = "Pause music"
 [[music.pause.example]]
 phrase = "pause spotify"
 test = true
-expectParamters = {service = "spotify"}
+expectParameters = {service = "spotify"}
 
 [[music.pause.example]]
 phrase = "stop"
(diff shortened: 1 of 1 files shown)


ianb:
It looks like this fixes some tests that were misspelled, but it doesn't change how the extension works. Is the underlying bug not present? (It's possible that the bug disappeared somewhere, but I want to make sure that's the case.)

melvin2016:
Yes, I have reviewed and tested the commands just like @alexandra-martin mentioned in #422 .

PR #1177, opened 2020-03-08
add what is playing intent

### LINKED ISSUE #1111: Intent: what's playing
This would look at the tab currently making sound, and maybe give its title, or a screenshot, or... something.

Question: does Spotify and other services set the title based on the currently playing song? If not we might have to add something to the services framework to get the currently playing item. We should still fall back to title and/or screenshot.

You can use `browser.tabs.query()` to find the tab(s) that are making sound.

annlinros:
@ianb  

Fixes #1111 

since Youtube and Spotify displays currently playing songs as title, extra services weren't required.
If two songs  plays in two tabs, the first one will be logged out.

this match - what song/music/title is playing - works
this match - what's playing & what is playing - doesn't work. 
can't figure out why.

### DIFF
--- a/extension/intents/music/music.js
+++ b/extension/intents/music/music.js
@@ -116,7 +116,7 @@ intentRunner.registerIntent({
 });
 
 intentRunner.registerIntent({
-  name: "music.find",
+  name: "music.showTitle",
   async run(context) {
     const tabs = await browser.tabs.query({ audible: true });
     if (!tabs.length) {
--- a/extension/intents/music/music.js
+++ b/extension/intents/music/music.js
@@ -115,5 +115,18 @@ intentRunner.registerIntent({
   },
 });
 
+intentRunner.registerIntent({
+  name: "music.find",
+  async run(context) {
+    const tabs = await browser.tabs.query({ audible: true });
+    if (!tabs.length) {
+      const e = new Error("Nothing is playing");
+      e.displayMessage = "Nothing is playing";
+      throw e;
+    }
+    context.displayText(tabs[0].title);
+  },
+});
+
 // FIXME: workaround for a legacy module needing access to this function:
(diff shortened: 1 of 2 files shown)


extension/intents/music/music.toml
```diff
@@ -143,3 +143,12 @@ phrase = "play music"
 [[music.unpause.example]]
 phrase = "play spotify"
 test = true
+
+[music.find]
+ description = "Find the tab that's currently making sound"
+match = """
+  (what | what's | which ) ( tab | song | music | title ) ( is ) playing
```
ianb: `( is )` means the same as `is`. I think it's supposed to be an optional word though, so you need to use `|` to say that no-word is an option, like `(is | )`

Also I think tab/song/music/title could be optional too.

extension/intents/music/music.js
```diff
@@ -115,5 +115,18 @@ intentRunner.registerIntent({
   },
 });
 
+intentRunner.registerIntent({
+  name: "music.find",
```
ianb: `music.find` doesn't seem like a good description, it would make me think of searching or focusing a tab. `music.showTitle` would be clearer.

extension/intents/music/music.toml
```diff
@@ -143,3 +143,12 @@ phrase = "play music"
 [[music.unpause.example]]
 phrase = "play spotify"
 test = true
+
+[music.find]
+ description = "Find the tab that's currently making sound"
```
ianb: Some leading whitespace here, but also we're not finding the tab as much as we're showing the title of the tab.

annlinros:
@ianb  I have made the changes. Please review.

ianb:
Looks good, thank you!

alexandra-martin:
Verified on Mac 10.14 and Win10x64 with Firefox Nightly 76.0a1 (2020-03-10).

Issue #557, opened 2019-11-11
Error displayed in "Doorhanger" when "Play" command doesn't find audio to play


alexandra-martin:
### Prerequisites:
Mic permissions are enabled.
A Spotify account is signed in and autoplay is enabled beforehand.

### STR:
1. Use the shortcut or click on the mic icon from the browser toolbar.
2. Say or write a Spotify "Play" command that will not have any results or any audio to play, e.g. "Play Ben Folds in Spotify" or "Play testtesttesttest".

### Expected result:
"Doorhanger" displays a message that audio cannot be played.

### Actual result:
An "Internal error: Error: Timeout waiting for #searchPage button[aria-label='Play'], .tracklist .tracklist-play-pause" error is displayed in the "Doorhanger".

### Notes:
Reproduced on Mac 10.14.6 and Win10x64 with Firefox Nightly 72.0a1 (64-bit).

![6](https://user-images.githubusercontent.com/54401091/68588050-3b4ea400-0491-11ea-8fc3-3e38a4d1d99d.gif)

ianb:
This is a bit complicated for a good first issue, but should be doable.

Mostly the content script (`extension/services/spotify/player.js`) should detect this not-found case, and raise an appropriate exception, then `spotify.js` should make that into an error with a `.displayMessage = "No results found"`.

JanviMahajan14:
@ianb I would like to work on this issue

noi5e:
@ianb with Outreachy, i'd like to give this a shot. can you assign this one to me?

noi5e:
@ianb @alexandra-martin I've got the dev environment set up, and I'm reproducing this bug. 

would love to work on it, if it's still a relevant contribution. just let me know!

alexandra-martin:
Verified on Mac 10.14 and Win 10x64 with Firefox Nightly 76.0a1 (2020-03-18).

![7](https://user-images.githubusercontent.com/54401091/76971607-12a25380-6936-11ea-83c1-c90747c9254a.gif)

PR #1178, opened 2020-03-08
Throw exception when no results are found for Spotify "search then play" intent

noi5e:
First Draft PR for issue [#557](https://github.com/mozilla/firefox-voice/issues/557):

pull_request

- Instead of "Timeout waiting for #searchButton," Voice now displays "No search results found"
- In player.js, I put the search for play button in a try-catch block, and threw the exception when it times out.
- This isn't quite what @ianb was suggesting (particularly having spotify.js handle displayMessage for the error). I spent a couple hours reading and tinkering, but still don't understand how to get spotify.js to handle an exception thrown in player.js...
- In spotify.js, try-catch won't catch a timeout error, and I can't find where console.log outputs.
- Help would be appreciated!


extension/services/spotify/player.js
```diff
@@ -18,12 +18,21 @@ this.player = (function() {
...
+            timeout: 2000,
+            // There seem to be 3 fixed buttons that appear early before the search results
+            minCount: 4,
+          });
+          playerButton.click();
+        } catch (e) {
+          if (e.name === "TimeoutError") {
+            const e = new Error("No search results");
+            e.name = "EmptyResults";
+            e.displayMessage = `No results found for ${query}`;
```
ianb: Oh... I think I pointed you the wrong way here. The exception is getting back, but `displayMessage` is getting lost. And right now we don't have a good way to keep it when the exception is passed from the content script (where `player.js` runs) to the background page (where `spotify.js` is running).

I'd recommend just throwing `new Error("No search results")` here and then catch it in `spotify.js`: https://github.com/mozilla/firefox-voice/blob/master/extension/services/spotify/spotify.js#L9 and add `.displayMessage` to it there. Something like:

```js
try {
  await this.callTab(...)
} catch (e) {
  if (e.message.includes("No search results")) {
    e.displayMessage = `...`
  }
  throw e;
}
```

noi5e:
@ianb delayed a bit by the craziness in the world, but i took the feedback you suggested and implemented it. voice now displays an error message with the query, saying no results were found.

pull_request

ianb:
@noi5e These are weird times indeed. I don't know if it's the perfect time to hunker down with some code, or if this is the time to look for more concrete things to do to be useful...

But, those thoughts aside, the PR is looking good ;)

PR #1180, opened 2020-03-08
Navigate to next and previous search result

### LINKED ISSUE #864: "next result" fails
**Issue Type:** Doesn't behave as expected
**Firefox Version:** Mozilla/5.0 (Windows NT 10.0; Win64; x64; rv:72.0) Gecko/20100101 Firefox/72.0
**Device:** Windows 10 PC
**Location:** UNITED STATES
**GitHub Username**: @SeijiSuenaga
## Steps to Reproduce
1) Said "search for Firefox Voice" and received Google results.
2) Said "next result" as suggested in the "big list of what you can say".

## Expected Behavior
I'm not sure what it's supposed to do...

## Actual Behavior
Received an error ("Sorry, there was an issue") and was told I hadn't made a search.

JM-Mendez:
This pull request addresses an issue where the command "next result" would throw an error to the user and not navigate properly.

To fix this, I had to use the provided function `performSearch` since the native extension javascript api—`browser.search.search`—would not provide any information about the search results or the tab.

Next, I added the ability for the command "previous results" to navigate to the search page, if currently on the first result.

I did not add any tests, since I did not see any for this part of the codebase. But if you'd like some, I can add those as well.

Closes #864 

### DIFF
--- a/extension/intents/search/search.js
+++ b/extension/intents/search/search.js
@@ -195,6 +195,7 @@ function pollForCard(maxTime) {
 async function moveResult(context, step) {
   stopCardPoll();
   const { tabId, searchInfo } = await getSearchInfo();
+
   if (!searchInfo) {
     const e = new Error("No search made");
     e.displayMessage = "You haven't made a search";
@@ -208,12 +209,17 @@ async function moveResult(context, step) {
     await context.makeTabActive(tabId);
     return;
   }
-  if (!(searchInfo.index + step >= 0)) {
+
+  // Initial search results do not have an index property and we want
+  searchInfo.index =
+    searchInfo.index === undefined ? 0 : searchInfo.index + step;
+
+  if (!(searchInfo.index >= 0)) {
     const e = new Error("No previous search result");
     e.displayMessage = "No previous search result";
     throw e;
   }
-  searchInfo.index += step;
+
   const item = searchInfo.searchResults[searchInfo.index];
   await browser.runtime.sendMessage({
     type: "showSearchResults",
--- a/extension/intents/search/search.js
+++ b/extension/intents/search/search.js
@@ -361,7 +361,12 @@ intentRunner.registerIntent({
   async run(context) {
     await performSearch(context.slots.query);
 
-    if (!buildSettings.android) await focusSearchTab();
+    if (!buildSettings.android) {
+      await focusSearchTab();
+
+      const searchInfo = await callScript({ type: "searchResultInfo" });
+      tabSearchResults.set(_searchTabId, searchInfo);
+    }
   },
 });
 
--- a/extension/intents/search/search.js
+++ b/extension/intents/search/search.js
@@ -201,24 +201,32 @@ async function moveResult(context, step) {
     e.displayMessage = "You haven't made a search";
     throw e;
   }
+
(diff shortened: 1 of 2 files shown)


JM-Mendez:
Hi @ianb, I noticed that I couldn't request a review for this PR. I saw that your preferred contact method was through video chat, so I'm assuming it's best to notify you through here for PRs?

extension/intents/search/search.js
```diff
@@ -359,12 +373,13 @@ intentRunner.registerIntent({
 intentRunner.registerIntent({
   name: "search.searchPage",
   async run(context) {
-    if (buildSettings.android) {
-      await performSearch(context.slots.query);
-    } else {
-      await browser.search.search({
-        query: context.slots.query,
-      });
+    await performSearch(context.slots.query);
```
ianb: I think this will effectively revert https://github.com/mozilla/firefox-voice/commit/10672d1f754e6d426571daade72135e6fe70dff0#diff-bcfc31c6e0fe90944fc2e9703795bbb1

That is, if you change your default search provider to DuckDuckGo, and say "search for ducks" you should end up at https://duckduckgo.com/?q=ducks&t=h_&ia=news

I'm not sure what the best answer here is. Maybe just fix it for Google, and use `performSearch()` if `googleIsDefaultProvider` is true.

ianb:
@JM-Mendez Thanks for the PR, I'm trying to regularly scan through them so I should pick up changes as they come in (though I'm getting a little behind with all the activity). You are right that there aren't any tests of what the intents actually do, we're just relying on manual testing.

JM-Mendez:
@ianb I managed to be able to use the default search engine for the query as you requested. I did so without having to use different queries for search engines besides google. It seems that I had to focus the tab in order to properly inject the content script.

One thing to note is that the tab is created hidden since I'm using the `openSearchTab` function. The first time the extension creates a tab it will request authorization from the user. Not sure if this is good UX for our use case though.

From mdn:
>The first time an extension hides a tab, the browser will tell the user that the tab is being hidden, show them how they can access the hidden tab, and give them the option of disabling the extension instead.
https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/tabs/hide

Currently the commands `previous result` and `next result` only retrieve the results for google searches. This happens due to how `communicate.register("searchResultInfo")` locates search results.

I'd like to create an issue for this and have it assigned to me if you don't mind.

At the moment I think I'll have to hardcode the html structure for different engines. Although I'll try to see if I can find some commonality I can use.
___
I also noticed a test failing locally, since tests are not run in CI. `git bisect` narrowed it down to commit 2a5f3d472d7e3dc761bf92a0d9afc0befcb56fa2. 

I can create an issue for this as well if you'd like. Doesn't have to be assigned to me though.

Here's a screenshot of the failing test.
Screen Shot 2020-03-10 at 13 44 38

ianb:
Cool, I think this is good. Of course the injected script doesn't work on other search engines, but you're working on resolving that so I think we can move forward. I mostly care about DuckDuckGo and Google, we haven't gotten user feedback about any other search engines.

The test failure you encountered has been fixed on master.

PR #1181, opened 2020-03-09
Implement combine all windows intent

### LINKED ISSUE #1103: Intent: combine all windows
This would move all tabs from other windows into the current window.

Suggested phrase: 

```
combine all windows (into |) (one |) (window |)
combine tabs (into |) (one |) window
```

fleur101:
Closes #1103 

### DIFF
--- a/extension/intents/window/window.js
+++ b/extension/intents/window/window.js
@@ -44,3 +44,13 @@ intentRunner.registerIntent({
     await browser.experiments.voice.openDownloads();
   },
 });
+
+intentRunner.registerIntent({
+  name: "window.combine",
+  async run(context) {
+    const currentWindow = await browser.windows.getCurrent();
+    const tabs = await browser.tabs.query({});
+    const tabsIds = tabs.map(tabInfo => tabInfo.id);
+    await browser.tabs.move(tabsIds, { windowId: currentWindow.id, index: -1 });
+  },
+});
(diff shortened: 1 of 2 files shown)


extension/intents/window/window.js
```diff
@@ -44,3 +44,13 @@ intentRunner.registerIntent({
...
   },
 });
+
+intentRunner.registerIntent({
+  name: "window.combine",
+  async run(context) {
+    const currentWindow = await browser.windows.getCurrent();
+    const tabs = await browser.tabs.query({});
+    const tabsIds = tabs.map(tabInfo => tabInfo.id);
+    await browser.tabs.move(tabsIds, { windowId: currentWindow.id, index: -1 });
```
ianb: I think this will also move tabs that are already in the current window, and might mix up the order of those tabs. So this should filter out tabs already in the current window.

extension/intents/window/window.toml
```diff
@@ -10,3 +10,19 @@ description = "Open downloads"
...
+"""
+
+[[window.combine.example]]
+phrase = "combine all windows"
+
+[[window.combine.example]]
+phrase = "combine all windows into one"
+
+[[window.combine.example]]
+phrase = "combine tabs into one window"
```
ianb: We only need one of these examples to actually display to the user, so the other examples should have `test = true` (they'll still be tested, but not displayed).

extension/intents/window/window.js
fleur101: Oh, yes, indeed. I added now filtering.

extension/intents/window/window.toml
fleur101: Okay, I see now! I also was not sure whether I need to add more examples or matchings and decided that this is sufficient

ianb:
Alright, looks good, thank you!

PR #1183, opened 2020-03-09
Close selected tabs intent

### LINKED ISSUE #1101: Intent: close selected tabs
To trace my thought process: I imagined a query "close [query] tabs", e.g., "close github tabs". It would do a search, probably only against the URL and title (not content, i.e., not the same as find.js), and then close those tabs. But closing all those tabs probably requires confirmation, and we don't have a pattern for that – including how to present the action, or adjust it.

Instead we could use the concept of selected tabs (partially exposed in Firefox, but easier to access through an extension). We could use "select [query] tabs" (or maybe "collect" as in #1086), and then "close selected tabs". The concept for "collect" both selects the tabs and collects them for easier review (a selected tab that's not visible in the tab bar is hard to review).

dave-ok:

Implements "Close selected tabs" intent. Fixes #1101

### DIFF
--- a/extension/intents/tabs/tabs.js
+++ b/extension/intents/tabs/tabs.js
@@ -371,3 +371,13 @@ intentRunner.registerIntent({
     await browser.tabs.move(tabsIds, { windowId: newWindow.id, index: 0 });
   },
 });
+
+intentRunner.registerIntent({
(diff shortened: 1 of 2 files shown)


ianb:
Looks good!

dave-ok:
Thanks!

PR #1185, opened 2020-03-09
Add progress indicator to Save Page

### LINKED ISSUE #792: Add progress indicator to Save Page
In #791 it adds a Save Page intent, but just putting together the HTML and resources and passing it back to the background page can take a second or two. During that time nothing appears to be happening. We should show some progress in the popup. 

Note that we can't do percentages, as we don't really know how long it will take.

We should probably have some message we send to the popup (by IntentContext) that shows a generic loading throbber, and we can use it other things in the future.

JanviMahajan14:

Fixes #792

### DIFF
--- a/extension/background/intentRunner.js
+++ b/extension/background/intentRunner.js
@@ -69,6 +69,19 @@ export class IntentContext {
     });
   }
 
+  savingPage(message) {
+    if (message === "startSavingPage") {
+      return browser.runtime.sendMessage({
+        type: "startSavingPage",
+        message,
+      });
+    }
+    return browser.runtime.sendMessage({
+      type: "endSavingPage",
+      message,
+    });
+  }
+
   displayText(message) {
     if (this.noPopup) {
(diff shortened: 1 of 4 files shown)


JanviMahajan14:
@ianb Can you check "Save Page" again . This is actually working in my system.

JanviMahajan14:
The only thing which i think is missing is that animation got hanged after some seconds. Is there some way i could get a infinite loop in the animation.

ianb:
@JanviMahajan14 Oh, you are right, I was missing the popup title change.

I think the animation isn't working because the page saving is CPU-intensive and it literally freezes the browser, including animations. The second time I tested it it at least got to the circle part of the animation, so I noticed it had changed, while before it seemed to freeze while it was still on the checkmark. I'm not sure exactly why that is.

Anyway, I think this is probably the best we can do for now given this animation.

Thanks!

JanviMahajan14:
Thank you so much @ianb

Issue #1089, opened 2020-02-26
Evaluate search libraries


ianb:
We're using [Fuse](https://fusejs.io/) but I don't know if it's the best option. [Flexsearch](https://github.com/nextapps-de/flexsearch) has some library comparisons.

For our uses we're creating somewhat small indexes on the fly and searching them once. Good search results and quick setup are most important to us.

PR #1186, opened 2020-03-09
Add collect tabs intent

### LINKED ISSUE #1086: Intent: Collect tabs
Maybe the simplest form would be "collect these tabs" which would get all tabs from the same domain and cluster them around the current tab. E.g., if I ran it right now it would take every github.com tab and move it around here (since I'm on the last tab, they would all be immediately before my active tab).

An extended form would be "collect all [query] tabs" which would do a search for the query and cluster all the matching tabs similarly. If the current tab wasn't matching, then it would... choose the best tab and also focus it...?

Finally "collect all tabs" would just rearrange all tabs according to domain (maybe sort by URL?).

And finally-finally, "collect all tabs in all windows" (or anything "in all windows") would grab tabs regardless of window.

danielamormocea:
Fixes #1086 

Added:
    1.  collect these tabs - collecting similar to active tab
    2.  collect [query] tabs - collecting tabs similar to query
    3. collect all tabs
    4. all of the above with "in all windows"

### DIFF
--- a/extension/intents/tabs/tabs.js
+++ b/extension/intents/tabs/tabs.js
@@ -544,3 +544,15 @@ intentRunner.registerIntent({
     }
   },
 });
+
+if (!buildSettings.android) {
+  intentRunner.registerIntent({
+    name: "tabs.closeSelectedTabs",
+    async run(context) {
+      const tabs = await browser.tabs.query({ highlighted: true });
+      const tabIds = tabs.map(tab => tab.id);
+
+      await browser.tabs.remove(tabIds);
+    },
+  });
+}
--- a/extension/intents/tabs/tabs.js
+++ b/extension/intents/tabs/tabs.js
@@ -371,3 +371,179 @@ intentRunner.registerIntent({
     await browser.tabs.move(tabsIds, { windowId: newWindow.id, index: 0 });
   },
 });
+
+async function collectTabsAround(center, tabs) {
+  if (tabs === undefined) {
+    return;
+  }
+
+  const tabsLeft = tabs.filter(tab => {
+    return tab.index < center.index;
+  });
+
+  const tabsRight = tabs.filter(tab => {
+    return tab.index > center.index;
+  });
+
+  for (let i = 0; i < tabsLeft.length; i++) {
+    await browser.tabs.move(tabsLeft[tabsLeft.length - i - 1].id, {
+      index: center.index - i - 1,
+    });
+  }
+
+  for (let i = 0; i < tabsRight.length; i++) {
+    await browser.tabs.move(tabsRight[i].id, { index: center.index + i + 1 });
+  }
+}
+
+async function collectRightTabs(tabs) {
+  if (tabs === undefined || tabs.length === 0) {
+    return;
+  }
+
+  const leftestTab = tabs[0];
+  const tabsRight = tabs.slice(1);
+
+  for (let i = 0; i < tabsRight.length; i++) {
+    await browser.tabs.move(tabsRight[i].id, {
+      index: leftestTab.index + i + 1,
+    });
+  }
+}
+
+function groupTabs(tabs, keyGetter) {
+  if (tabs === undefined) {
+    return {};
+  }
+
+  const groupedTabs = {};
+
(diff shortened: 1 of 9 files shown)


extension/intents/tabs/tabs.js
```diff
@@ -371,3 +371,179 @@ intentRunner.registerIntent({
     await browser.tabs.move(tabsIds, { windowId: newWindow.id, index: 0 });
   },
 });
+
+async function collectTabsAround(center, tabs) {
+  if (tabs === undefined) {
+    return;
+  }
```
ianb: What's the intention here? It seems like this should be an error

extension/intents/tabs/tabs.js
```diff
@@ -371,3 +371,179 @@ intentRunner.registerIntent({
...
+    const currentWindow = await browser.windows.getCurrent();
+    const allWindows = await browser.windows.getAll({
+      windowTypes: ["normal"],
+    });
+
+    const matchingTabs = await getMatchingTabs({
+      query: context.slots.query,
+      activeTab,
+      sort_by_index: true,
+      currentWindow: context.parameters.currentWindow,
```
ianb: Generally I would try to convert the string as soon as possible, so instead of testing `currentWindow === "true"` in the other function, I would do `currentWindow: context.parameters.currentWindow === "true"` right here

extension/intents/tabs/tabs.toml
```diff
@@ -262,3 +262,27 @@ match = """
 
 [[tabs.moveSelectedToNewWindow.example]]
 phrase = "Move selected tabs to new window"
+
+[tabs.collectAllTabs]
+description = "Collect all tabs by url"
+match = """
+  collect (all | the|) tabs in all (the|) windows [currentWindow=false]
```
ianb: I'd add `(in | from)` here, and use `window{s}` to make the pluralization optional.

extension/intents/tabs/tabs.js
```diff
@@ -371,3 +371,179 @@ intentRunner.registerIntent({
...
+    const activeTab = options.activeTab;
+    const activeHostname = new URL(activeTab.url).hostname;
+    tabQuery.url = "*://*." + activeHostname + "/*";
+  }
+
+  let matchingTabs = await browser.tabs.query(tabQuery);
+
+  if (options.query !== undefined) {
+    matchingTabs = matchingTabs.filter(tab => {
+      return new URL(tab.url).origin.includes(options.query);
```
ianb: This should also test the title, and we'll want a case-insensitive test. Arguably we should be using Fuse for searching (or some other library per #1089), but I think for now a simple string search is OK and adding fuzzy search can be a followup ticket.

extension/intents/tabs/tabs.toml
```diff
@@ -262,3 +262,27 @@ match = """
...
+"""
+
+[[tabs.collectAllTabs.example]]
+phrase = "Collect all tabs"
+
+[tabs.collectMentionedTabs]
+description = "Collect tabs similar to active tab or by query"
+match = """
+  collect these tabs [currentWindow=true]
+  collect these tabs in all (the|) windows [currentWindow=false]
```
ianb: Also I think this could match `collect tabs (like | from |) (this |) (site | page | url | domain | host)` and then another version from all windows.

And (in | from) here as well

danielamormocea:
@ianb I made the requested changes. Could you please review?

extension/intents/tabs/tabs.js
```diff
@@ -372,6 +372,178 @@ intentRunner.registerIntent({
...
+    tabQuery.url = "*://*." + activeHostname + "/*";
+  }
+
+  let matchingTabs = await browser.tabs.query(tabQuery);
+
+  if (options.query !== undefined) {
+    matchingTabs = matchingTabs.filter(tab => {
+      return (
+        new URL(tab.url).origin.includes(options.query) ||
+        tab.title.toLowerCase().includes(options.query)
```
ianb: The query can also be upper case, so you should also make sure to convert it to lower case as well.

ianb:
Great, thank you!

PR #1192, opened 2020-03-09
Fixes #912 - Support sound cloud

### LINKED ISSUE #912: Support Soundcloud
Based on telemetry users were hoping audio controls would work on SoundCloud.

"Play"
"Pause"
"Mute"

melvin2016:

Fixes #912 .
 
Added:
1. Play
2. Pause/Stop
3. Previous track
4. Skip track

`npm test` is getting failed locally.
@ianb Sir, can you review this and add some suggestions?


extension/services/soundcloud/player.js
```diff
@@ -0,0 +1,49 @@
+/* globals helpers */
+
+this.player = (function() {
+  // Play button Reference
+  const SEARCH_PLAY = ".sound__header a[title='Play']";
+
+  class Player extends helpers.Runner {
+    action_play() {
+      const button = this.querySelector("button[aria-label='Play current']");
```
ianb: Ideally these would use selectors that didn't use these English phrases, as they will break in other languages. There's several open bugs in Spotify for the same issues. (Of course I can see why you'd select these, because I did offer up Spotify as a good example...)

If it's not easy to figure out other selectors we can do a followup bug to fix these.

extension/services/soundcloud/soundcloud.js
```diff
@@ -0,0 +1,74 @@
...
+        }
+      }
+    }
+  }
+
+  async move(direction) {
+    const tabs = await this.getAllTabs();
+    if (!tabs.length) {
+      const e = new Error("Spotify is not open");
+      e.displayMessage = "Spotify is not open";
```
ianb: You should switch the name here

extension/services/soundcloud/soundcloud.js
```diff
@@ -0,0 +1,74 @@
+import * as music from "../../intents/music/music.js";
+import * as serviceList from "../../background/serviceList.js";
+import * as content from "../../background/content.js";
+import { shouldDisplayWarning } from "../../limiter.js";
+
+class Soundcloud extends serviceList.Service {
```
ianb: It's possible we should be making an abstract base class that both Spotify and Soundcloud (and future services) inherit from.

melvin2016: I have added an abstract class of MusicService but I don't know if this is what you meant. If not, can you give some pointer how can I do that?

extension/services/soundcloud/soundcloud.js
melvin2016: Done

extension/services/soundcloud/player.js
melvin2016: removed selectors which used English phrases and added selectors based on class names.

melvin2016:
@ianb Please review this.

extension/services/soundcloud/soundcloud.js
ianb: Ah, cool, yes, what you did is what I meant

melvin2016:
@ianb Thank You!

PR #1195, opened 2020-03-09
1059 clean up url

annlinros:
CleanUrl function cleans up all google analytics params as of now.

when I test cleanURL("url") alone it works.
Example: 
`https://www.example.com/?utm_source=summer-mailer&utm_medium=email&utm_campaign=summer-sale`
=>
`https://www.example.com/`

I have some questions:
1. Should I include any other params to be filtered out ?
2.I get an error from ESLint after testing 
` "error  Function 'cleanURL' expected no return value  consistent-return"`
But according to my understanding, `browser.webRequest.onBeforeRequest.addListener( )` must have a function as the first argument. Is there a work-around ?

Thanks.


extension/background/pageMetadata-contentScript.js
```diff
@@ -41,5 +41,105 @@ this.pageMetadataContentScript = (function() {
...
+  function cleanURL(details) {
+    const globalBlockedParams = [
+      "utm_source",
+      "utm_medium",
+      "utm_campaign",
+      "utm_term",
+      "utm_content",
+    ];
+
+    function getParams(URL) {
```
ianb: Instead of all this string-based splitting you can use the `URL` object, kind of like:

```js
const u = new URL(url);
const params = u.searchParams;
params.delete("utm_source");
const betterUrl = u.toString();
```

Obviously doing the other cleanup too. You can read about searchParams here: https://developer.mozilla.org/en-US/docs/Web/API/URL/searchParams

The cool thing about URL objects is you can edit them and then turn them back into strings.

extension/background/pageMetadata-contentScript.js
```diff
@@ -41,5 +41,105 @@ this.pageMetadataContentScript = (function() {
...
+    }
+
+    const domain = getDomain(details);
+    if (domain === null) {
+      return;
+    }
+
+    const blockedParams = [];
+    for (const gbp of globalBlockedParams) {
+      if (!gbp.includes("@")) {
```
ianb: I don't understand what this is doing, none of `globalBlockedParams` seem to have `@` in them?

extension/background/pageMetadata.js
```diff
@@ -14,3 +14,12 @@ export async function getMetadata(tabId) {
...
   });
 }
+
+export async function cleanURL(url) {
+  await content.lazyInject(url, "/background/pageMetadata-contentScript.js");
+  browser.webRequest.onBeforeRequest.addListener(
+    cleanURL,
+    { urls: ["<all_urls>"] },
+    ["blocking"]
+  );
```
ianb: We don't need or want to actually change the URLs in the browser for this, just to change the URL for the purposes of copying the link. So if you go to a page with `?utm_campaign=whatever` and then say "copy link", that utm_campaign won't be in the copied link.

So all this code can actually live in `pageMetadata.js` and instead of this:

```js
  return browser.tabs.sendMessage(tabId, {
    type: "getMetadata",
  });
```

it would be like:

```js
  const metadata = await browser.tabs.sendMessage(tabId, {
    type: "getMetadata",
  });
  metadata.canonical = cleanUrl(metadata.canonical);
  return metadata;
```

annlinros:
@ianb  I have refactored the cleanURL function using URL() Objects. 
Please review.

extension/background/pageMetadata.js
```diff
@@ -10,7 +10,43 @@ export async function getSelection(tabId) {
...
+  function buildURL(baseURL, params) {
+    if (Object.keys(params).length === 0) {
+      return baseURL;
+    }
+    const newURL = baseURL + "?" + params;
+    return newURL;
+  }
+
+  const baseURL = string.split("?")[0];
+  const params = getReducedParams(string);
```
ianb: You can actually remove most of this. I'd rename `getReducedParams()` to `reduceParams()` and then have it return `url.toString()`. It will remove the parameters and rebuild the URL all on its own (`urlObject.toString()` rebuilds the URL)

annlinros:
@ianb  Please review.

ianb:
This looks good. Thank you!

PR #1198, opened 2020-03-09
fixed error in english.toml

saucekode:


extension/background/language/langs/english.toml
```diff
@@ -18,6 +18,8 @@ haste = "paste"
 taste = "paste"
 pace = "paste"
 best = "paste"
+"copy links" = "copy"
+coffee = "copy"
```
ianb: Misheard words are perfect for this file, so yes to this one

extension/background/language/langs/english.toml
```diff
@@ -18,6 +18,8 @@ haste = "paste"
 taste = "paste"
 pace = "paste"
 best = "paste"
+"copy links" = "copy"
```
ianb: "copy links" doesn't _always_ mean exactly the same as "copy". E.g., "copy links title" shouldn't mean the same thing as "copy title" (I don't know what it should mean... but it doesn't make much sense).

So instead this should be changed in `clipboard.toml` so that the existing `copy link` phrase also matches `copy links`

extension/views/lexicon.html.ejs
```diff
@@ -260,6 +260,7 @@
             Example
             <p>Copy (copies whatever is selected)</p>
             <p>Copy link</p>
+            <p>Copy links</p>
```
ianb: We don't need to tell users about this specific phrasing, it's just there so that if it happens to be transcribed as "copy links" it will still work. So this can be removed.

extension/intents/clipboard/clipboard.toml
```diff
@@ -21,12 +21,15 @@ test = true
...
-  copy (the |) (this |) (tab's |) (link | url) (of this tab |) (to clipboard |)
+  copy (the |) (this |) (tab's |) (link{s} | url) (of this tab |) (to clipboard |)
 """
 
 [[clipboard.copyLink.example]]
 phrase = "Copy link"
 
+[[clipboard.copyLink.example]]
+phrase = "Copy links"
+
```
ianb: This could be removed or marked as `test = true`

ianb:
Trying to clean up the PR backlog, will fix lexicon in master...

PR #1199, opened 2020-03-09
Fixes #1082 - Select tab by number

### LINKED ISSUE #1082: Select tab by number
We have next/previous/first/last tabs (implemented in #1078). We could also support numbers, like "go to tab 23" or "go to second tab".

It would probably be good to extend the `smallNumber` type in `entityTypes.js` (in a new type), and include cardinals (first, second, third, etc).

sabina-rohman:
1. Added new intent for going to the Nth tab
2. N can be a number, ordinal or cardinal
3. Store the numbers/cardinals in a separate file numbers.js
4. Refactored smallNumbers entityType to reuse these numbers


extension/background/numbers.js
```diff
@@ -0,0 +1,112 @@
...
+  "12th",
+  "13th",
+  "14th",
+  "15th",
+  "16th",
+  "17th",
+  "18th",
+  "19th",
+  "20th",
+];
```
ianb: As a followup it would be good to move these into `english.toml`

extension/background/numbers.js
```diff
@@ -0,0 +1,112 @@
...
+  } else if (ordinalIndex !== -1) {
+    return ordinalIndex;
+  } else if (cardinalIndex !== -1) {
+    return cardinalIndex;
+  } else if (numberedOrdinalsIndex !== -1) {
+    return numberedOrdinalsIndex;
+  } else {
+    return -1;
+  }
+}
```
ianb: And if the words go in english.toml, this would go in `lang.js` (as a method of `Language`)

extension/background/numbers.js
```diff
@@ -0,0 +1,112 @@
...
+  if (numberIndex !== -1) {
+    return numberIndex;
+  } else if (ordinalIndex !== -1) {
+    return ordinalIndex;
+  } else if (cardinalIndex !== -1) {
+    return cardinalIndex;
+  } else if (numberedOrdinalsIndex !== -1) {
+    return numberedOrdinalsIndex;
+  } else {
+    return -1;
```
ianb: I've always disliked `.indexOf()` because if you don't check the return value you get this crazy number slipping in, and there's also more than one way to check the return value.

Anyway, I think `undefined` or `null` is a better return value here.

extension/background/numbers.js
```diff
@@ -0,0 +1,112 @@
...
+  "20th",
+];
+
+export function findMatchingIndex(itemName) {
+  //Return the index of the itemName from the any of the number arrays that contains it
+  const val = itemName.toLowerCase();
+  const numberIndex = numbers.indexOf(val);
+  const ordinalIndex = ordinals.indexOf(val);
+  const cardinalIndex = cardinals.indexOf(val);
+  const numberedOrdinalsIndex = numberedOrdinals.indexOf(val);
```
ianb: Are these correct? It seems like "first" is going to return 0, "second" 1 and so on.

Though it's a little more tedious, `"1st": 1, "first": 1, "one": 1, "2nd": 2, ...` might be a better way to define these. Then we could add something like `"1th": 1` later if we somehow see people accidentally saying that (i.e., not assume there's only one way to say each kind of number).

extension/intents/tabs/tabs.js
```diff
@@ -556,3 +557,21 @@ if (!buildSettings.android) {
...
+
+intentRunner.registerIntent({
+  name: "tabs.gotoTabNumber",
+  async run(context) {
+    const userInput = context.slots.tabNumber;
+    const tabNumber = findMatchingIndex(userInput);
+
+    if (tabNumber < 0) {
+      const erroMessage = userInput + " is not a vliad tab number";
+      const e = new Error(erroMessage);
```
ianb: A couple typos here: erroMessage, and vliad.

Also, the error should always be a static string, like `new Error("Invalid tab number")`. Since the error message is our only way to categorize errors, user input in the error message will make that harder.

`displayMessage` though can contain user input. I'd recommend using something like:

```js
e.displayMessage = `"${userInput}" is not a valid number`;
```

String interpolation tends to be a bit clearer in my experience.

Issue #1182, opened 2020-03-09
Update readme with clearer instructions for linux.


LilyLME:
The developing part of the readme starts of well by asking that nodejs is downloaded then suddenly moves to **_"After downloading Firefox Nightly, change its name to FirefoxNightly (no space), otherwise the project wont be able to spawn the firefox."_**  That instruction is a little bit confusing we might want to restructure it by:

First asking the that the reader downloads Firefox nigtly from : https://www.mozilla.org/en-US/firefox/channel/desktop/

Next the instruction about  changing _**"its name"**_  is not clear, what is being renamed? The binary `fireforx`? that's not clear.

The ambiguous instruction can be found at :
https://github.com/mozilla/firefox-voice/blame/master/README.md#L36

I had a work around by applying these instructions from the WSL sections, otherwise I was quite lost.
https://github.com/mozilla/firefox-voice/blame/master/README.md#L80-L81

Hope I did not miss something :)

av-00:
I've just completed setting up everything after a lot of support, @ianb can I work on it?

ianb:
@av-00 Yes, that would be great, thanks!

PR #1201, opened 2020-03-10
Update README.md

av-00:
Issue #1182 . Please tell if if I made any errors or if it needs improvement. @ianb

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,6 +34,7 @@ Note:
 
 1. Download the latest LTS version of NodeJs before running above npm commands.
 2. After downloading Firefox Nightly, change its name to FirefoxNightly (no space), otherwise the project wont be able to spawn the firefox.
+3. Clone firefox-voice to your system and enter the directory.
 
 The developer installation is:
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -46,7 +47,7 @@ This will launch a new Firefox browser with the extension installed. You should
 
 You may face errors on performing ```npm install``` that can be resolved by updating the node to its latest version [see here](https://www.hostingadvice.com/how-to/update-node-js-latest-version/)
 
-If a new browser does not open, it might be because the path to Nightly is not found. Use the command `FIREFOX="/usr/bin/firefox" npm start` instead.
+If a new browser does not open, it might be because the path to Nightly is not found(most probable case). Find your binary file in the system, and enter the entire path to it. Binary file is the application through which firefox nightly can be opened manually. Use the command `FIREFOX="/usr/bin/firefox" npm start` instead.
```
ianb: Oops, you are missing a space before the parenthesis. It might be clearer if you say "Find your application's executable: the file where Firefox Nightly can be opened manually. Enter the entire path of the file"

PR #1209, opened 2020-03-10
fix: open bookmark opens wrong bookmark

Simpcyclassy:


Simpcyclassy:
@ianb  please could you review? Thanks

ianb:
Hmm... I wouldn't expect the words would need to be split. But I can't reproduce the original bug so I can't figure out if this fixes it or not. I'm going to land this, @alexandra-martin can you try to verify the fix?

PR #1210, opened 2020-03-10
Set minimum node version

naima-shk:
My 'npm test' command showing some errors  into bin/calculate-version.js .
@ianb  Kindly review :)

### DIFF
--- a/package.json
+++ b/package.json
@@ -3,10 +3,10 @@
   "id": "firefox-voice@mozilla.org",
   "description": "This is a WebExtension that allows the browser to be voice-controlled.",
   "private": true,
+  "version": "0.19.0",
   "engines": {
-    "node": ">=12.16.1"
+    "node": ">=12.0.0"
   },
-  "engineStrict": true,
   "author": {
     "name": "Mozilla Emerging Technologies",
     "url": "https://github.com/mozilla/firefox-voice"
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -3,7 +3,10 @@
   "id": "firefox-voice@mozilla.org",
   "description": "This is a WebExtension that allows the browser to be voice-controlled.",
   "private": true,
-  "version": "0.19.0",
```
GangaChatrvedi: why removed this line?

package.json
```diff
@@ -3,7 +3,10 @@
...
+  "engines": {
+    "node": ">=12.16.1"
+  },
+  "engineStrict": true,
```
GangaChatrvedi: Generally **engineStrict** is used if we have clear idea about the least version to use. **12.16.1** is a bit high version for that.

According me engineStrict should be avoided in our case.

@ianb am i right?

GangaChatrvedi:
Also putting us good PR title helps.

Try something like 

**Set minimum node version in package.json**

package.json
ianb: This is also what's causing the build to fail: https://circleci.com/gh/mozilla/firefox-voice/2588

package.json
ianb: Probably we can just say `>=12.0.0` – at least unless we test it out and are sure that there's something in a later version that we require. 12.16.1 is very exact, so it _seems_ like we're being specific, but I'm assuming this just happens to be the latest 12 release.

I wasn't familiar with engineStrict, but looking it up it seems the property is deprecated: https://docs.npmjs.com/files/package.json#enginestrict

naima-shk:
@ianb  Please update review.

PR #1216, opened 2020-03-10
Add alias so 'book mark' means 'bookmark'

ianb:


ianb:
Good work Ian, thank you for your contribution!

Issue #1042, opened 2020-02-20
Match phrase "search"


ianb:
If you just say "search" (no query) it acts kind of weird.

It should open your default search engine. This can be done with `browser.search.search({query: ""})`

This should be a new intent in `search.js`

tonynguyen111997:
Hi @ianb , I would like to take on this issue.

ianb:
@tonynguyen111997 You got it!

PR #1223, opened 2020-03-10
Match phrase search

tonynguyen111997:
This PR references issue #1042 

Changes: 
`search.toml`: line 125 - 133
Added a new intent in `search.toml` called `[search.defaultSearchEngine]`

`search.js`: line 274 - 279
Registered new intent in `search.js` that opens default search engine when user
says "search" with nothing after


extension/intents/search/search.toml
```diff
@@ -121,3 +121,13 @@ test = true
...
+
+[search.defaultSearchEngine]
+description = "Opens the default search engine"
+match = """
+  search
+"""
+
+[[search.defaultSearchEngine.example]]
+phrase = "search"
+test = true
```
espertus: Could you please end the file with a newline? (GitHub highlights missing newlines in PRs, which is how I caught it.)

tonynguyen111997: Of course! Let me go ahead and add it. Thank you for pointing it out.

tonynguyen111997: @espertus I just made the changes. 😄 I hope you don't mind me asking, but I'm curious as to why new lines are needed at the end of the file? Can you please explain this to me or point me towards where I can learn more about this?

espertus:
Some tools don't work write if the input file doesn't end with a newline.
Many older programmers have been burnt by this in the past.

https://unix.stackexchange.com/questions/18743/whats-the-point-in-adding-a-new-line-to-the-end-of-a-file

https://stackoverflow.com/questions/729692/why-should-text-files-end-with-a-newline

extension/intents/search/search.toml
ianb: In addition to possible (if unlikely) problems, having a newline at the end of the file avoids unimportant diffs later if someone else edits the file. For the same reason it's good to avoid trailing spaces on a line.

You can probably configure your editor to both trim trailing whitespace, and to make sure there's one and only one newline at the end of a file (e.g., in VS Code it's Insert Final Newline and Trim Final Newlines). Neither changes the meaning of the code, which is why keeping them clean avoids future changes-that-aren't-really-changes.

Issue #1197, opened 2020-03-09
Allow nicknames to be edited in options


ianb:
Right now we have a bunch of nicknames managed in `nicknames.js`. But you have to use voice commands, and it's hard to do that precisely or understand what you've saved.

I imagine a new section at the bottom of the settings page, looking roughly like:

<img width="516" alt="image" src="https://user-images.githubusercontent.com/44390/76261739-a8e4c400-6228-11ea-85b5-881e30cf28cd.png">

The "Save" button might not be anything special, most of those fields in settings save when they are changed or unfocused. Delete of course deletes the one routine. I'm borrowing the name "routine" from Alexa, instead of calling them nicknames.

Note that you'll have to get the existing routines from `intentRunner.js` and then also save them there. This will require some additional messages being sent from `optionsController.jsx` to `main.js` through the `intentRunner.js`, both to fetch and save nicknames. 

The code currently distinguishes between a single alias, and a nickname that does several actions (`combined`), but the UI shouldn't show them differently.

@awallin might have some design notes, but I think the simple UI is enough to move forward with.

danielamormocea:
@ianb. ready to take this :D

awallin:
I can definitely make a higher fidelity UI design. My initial thought is this shouldn't be on the main settings page but instead linked from there. The settings page is getting long as is and it's unlikely this feature will be used by a majority of users.

ianb:
@awallin Can we make a tabbed version of the settings page? Just in terms of code it seems nice to keep this stuff in one page.

awallin:
@ianb Similar to the Firefox Preferences page which creates section for General, Home, Search, Privacy & Security, and Sync?

danielamormocea:
@ianb @awallin So should I make a simple UI alongside the backend functionalities, then get feedback and improve the UI later?

ianb:
@danielamormocea Yes, I generally like to focus on just having each step improve things. So as long as the UI flow seems right we can rearrange or restyle things in followups.

awallin:
Here's a pass at the designs for routines including tabbed sidebar and layout for routines on a separate in-page tab. There's more to come for editing a specific routine and confirmations for delete.

Let me know if you have any questions?

![preferences-with-navigation](https://user-images.githubusercontent.com/428245/76449733-d301c780-6389-11ea-9d71-41b329d8b4db.jpg)

[...]

PR #1225, opened 2020-03-11
added tab navigation

LilyLME:
This PR adds tabs to the settings page as suggested in #1197 

Signed-off-by: LilyLME <brilyjoy@gmail.com>


LilyLME:
I added the navigation buttons alone just we can follow up if we would want it to look this way, if so I would then proceed to binding the tabs with the nav buttons.

tabs design

LilyLME:
@ianb @awallin Pls I would need some guidance here and suggestions how to go about it further. Hacking my way through React for the first time 🙈

LilyLME:
Looks like this now.

options-tab-demo

extension/options/optionsView.jsx
```diff
@@ -10,31 +10,36 @@ export const Options = ({
...
+const LeftSidebar = ({ version, updateActiveTab }) => {
   return (
     <div className="settings-sidebar">
       <img src="./images/firefox-voice-stacked.svg" alt="Firefox Voice Logo" />
+      <Tabs updateActiveTab={updateActiveTab} />
       <div className="version-info">
         <p>Version {version}</p>
         <p>
-          <a href="/views/CHANGELOG.html">What's New</a>
+          <a href="/views/CHANGELOG.html">What"s New</a>
```
ianb: this looks unintentional

extension/options/optionsView.jsx
```diff
@@ -498,3 +557,49 @@ const AboutSection = () => {
     </fieldset>
   );
 };
+
+const Tabs = ({
+  updateActiveTab
+}) => {
+  return (
+    <fieldset id="tabs">
```
ianb: A fieldset isn't necessary here. This isn't really a form, it's just navigation.

extension/options/optionsView.jsx
```diff
@@ -498,3 +557,49 @@ const AboutSection = () => {
...
+      <ul>
+        <li>
+          <button className="styled-button" onClick={() => updateActiveTab("chime-settings")}>
+            Preferences
+          </button>
```
ianb: What I'd actually prefer here is to just use:

```html
<a className="tab-button" href="#chime-settings">Preferences</a>
```

And then use [onhashchange](https://developer.mozilla.org/en-US/docs/Web/API/WindowEventHandlers/onhashchange) to catch the change. This way the URL reflects the location, the back button will work, you can bookmark it, etc. Also the hash should be checked on page load.

One thing I'm a little unhappy with in the current code is that `OptionsController` uses state internally, and it's not easy to get state in from the outside (or maybe I just don't understand how to do it). But adding a hashchange handler inside the function isn't good, since the function can (I believe) get called multiple times as state changes.

An example of the solution I came up with is `onKeyboardShortcutError` – it's not a great solution, but at least if you use the same solution then they will be not-great in the same way.

LilyLME: Thanks @ianb I would add the mentioned changes.

extension/options/optionsView.jsx
LilyLME: oops, yep, would get that.

LilyLME:
Previously requested changes and updates included. :)

ianb:
Great, this is looking good. There are some style issues... but there's also an issue with this PR overlapping with another one.

I think the best way to approach this is to land this PR, and then I will have to revert it to apply the other PR. But that way your commits will be part of your github record and the history of this project.

PR #1226, opened 2020-03-11
Add in files from espertus/mozilla-assistant

espertus:
I ran the following:


android-app/app/src/main/assets/langs/english.toml
```diff
@@ -0,0 +1,71 @@
+[aliases]
```
ianb: As a followup we'll want to load this file out of `extension/background/language/langs/english.toml`

PR #1229, opened 2020-03-11
scrolling on pdf pages

### LINKED ISSUE #954: Scrolling (etc) doesn't work on a PDF page

danielamormocea:
Fixes #954 

This adds logic to distinguish between pdf pages and the others.
Don't want to interfere with the already assigned person.

### DIFF
--- a/extension/intents/scroll/scrollHelper.js
+++ b/extension/intents/scroll/scrollHelper.js
@@ -2,14 +2,28 @@
 
 const scrollAmount = 0.9;
 
-function scrollVertically(dy, smooth) {
-  let toBeScrolled = document.activeElement;
+function getScrollableElement() {
+  let element = document.activeElement;
+  let height = element.offsetHeight;
+  let scrollPos = element.scrollTop;
+  const scrollHeight = element.scrollHeight;
 
   if (document.contentType === "text/html") {
-    toBeScrolled = window;
+    element = window;
+    height = element.innerHeight;
+    scrollPos = element.scrollY;
   }
 
-  toBeScrolled.scrollBy({
+  return {
+    element,
+    height,
+    scrollPos,
+    scrollHeight
+  };
+}
+
+function scrollVertically(dy, smooth, element) {
+  element.scrollBy({
     left: 0,
     top: dy,
     behavior: smooth ? "smooth" : "auto",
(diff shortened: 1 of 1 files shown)


extension/intents/scroll/scrollHelper.js
```diff
@@ -3,7 +3,13 @@
 const scrollAmount = 0.9;
 
 function scrollVertically(dy, smooth) {
-  window.scrollBy({
+  let toBeScrolled = window;
+
+  if (document.contentType === "application/pdf") {
```
GangaChatrvedi: Really elegant way to solve the problem.  

My only thought is. There can be many other contents as well that can be covered by your code (in addition to pdf). What if we put add html related contentType with **window** and ```toBeScrolled = document.activeElement``` for remaining.

danielamormocea:
@GangaChatrvedi  made the requested change, if you can take a look :)

danielamormocea:
@ianb @GangaChatrvedi  Could you please take a look at the changes above?

I forgot initially to make this work for scroll top and scroll bottom; I saw it later on the other PR.

extension/intents/scroll/scrollHelper.js
```diff
@@ -2,8 +2,28 @@
...
+    scrollPos = element.scrollY;
+  }
+
+  return {
+    element,
+    height,
+    scrollPos,
+    scrollHeight,
+  };
+}
```
ianb: While this works for the PDF example, it won't work for other pages, like a Bugzilla bug (https://bugzilla.mozilla.org/show_bug.cgi?id=343205) or a Google doc. I think the `document.contentType` check isn't the right thing: rather `document.activeElement` _or one of its ancestors_ is always the right thing to scroll.

This comment also gives some guidance: https://github.com/mozilla/firefox-voice/issues/954#issuecomment-591691095

ianb:
#1235 is also related to this

danielamormocea:
@ianb  I tried making this work for google docs, but I couldn't. It works for the rest of the sites well, but google has a weird object, and I can't figure it out :(

ianb:
Oh yeah, I just looked at Google Docs, it is very weird. Full of weird invisible iframes... I wonder if they use them to capture different input and then just create their own rendering system on top of that.

But yeah, looks good. Approved but holding for #1245

PR #1230, opened 2020-03-11
Add .nvmrc pointing toward latest LTS version of node

GangaChatrvedi:


ianb:
@GangaChatrvedi Does this solve something that `engines` in `package.json` doesn't?

GangaChatrvedi:
@ianb It is not actually about solving the problem. More about helping people who use **nvm** tool to shift to right version.

User do not have to know right version. They can simply use ```nvm use``` and they are ready to go with best version we suggest.

ianb:
OK... I suppose that's okay. I worry if someone has, say, 12.6.2 installed (whenever that comes out) that this might cause some issue, but we can see how it goes.

Issue #1211, opened 2020-03-10
Cursor should be pointer for smile/frown feedback


awallin:
The smile/frown icons should use a pointer (hand) when hovering on the smile/frown in the doorhanger. This indicates to a user they can click the icons.

nuraynab:
Hi, can I have this?

awallin:
Go for it @nuraynab

alexandra-martin:
Verified on Mac 10.14 and Win10x64 with Firefox Nightly 76.0a1 (2020-03-11).

![21](https://user-images.githubusercontent.com/54401091/76523590-54805500-6471-11ea-8051-c28a1c82660d.gif)

PR #1232, opened 2020-03-11
added pointer for smile/frown feedback

nuraynab:
This PR fixes issue #1211  

Added pointer cursor to feedback-controls in popup.css file

### DIFF
--- a/extension/popup/popup.css
+++ b/extension/popup/popup.css
@@ -307,18 +307,22 @@ a#lexicon {
 
 .feedback-controls .happy-icon {
   mask: url("./images/icon-happy.svg");
+  cursor: pointer;
 }
 
 .feedback-controls .sad-icon {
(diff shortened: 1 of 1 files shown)


extension/popup/popup.css
```diff
@@ -307,18 +307,22 @@ a#lexicon {
...
 }
 
 .feedback-controls .sad-icon {
   mask: url("./images/icon-sad.svg");
+  cursor: pointer;
 }
 
 .feedback-controls .happy-icon:hover {
   mask: url("./images/icon-happy-hover.svg");
+  cursor: pointer;
```
ianb: I don't think these will be necessary for `:hover`, they should inherit the cursor from the previous rules

nuraynab: Okay, I'll change it. Thanks!

Issue #1213, opened 2020-03-10
"Previous/Next results" commands only work with google search


JM-Mendez:
The commands `previous results` and `next results` are only able to fetch results from the google. This happens because the results are iterated over according to google's html structure, which doesn't match the other providers.

JM-Mendez:
@ianb can you please assign this issue to me? It's a continuation of PR #1180, but was not within scope of the issue that PR was addressing.

ianb:
@JM-Mendez Cool, you got it

ianb:
QA note: if you change your default search engine to DuckDuckGo, and then run a search, and say "next result", it should go to the next result (like it does for Google)

JM-Mendez:
> QA note: if you change your default search engine to DuckDuckGo, and then run a search, and say "next result", it should go to the next result (like it does for Google)

@ianb this should have been resolved when #1236 landed. Is this a regression I need to work on?

ianb:
@JM-Mendez: @alexandra-martin wanted to verify the fix but wasn't clear what needed to be verified.

12people:
@ianb @JM-Mendez isn't working for me in the extension (version 0.21.0).
Is this a regression?

PR #1236, opened 2020-03-11
Scrape search results from Bing and DuckDuckGo

JM-Mendez:
Currently the only search engine being scraped for results is Google. This PR scrapes from Bing and DuckDuckGo. 

I originally wanted to rely solely on the document structure, but code reusability would decrease by re-implementing logic for each engine. Since—unlike Google—Bing and DuckDuckGo use BEM to structure their classes, I felt this should be sufficient for now.

I also throw an error (screenshot below) if search results cannot be scraped, which instructs the user to inform us. This would help us address the moment the structure or classes change. For this I had to add the `white-space` rule to prevent newlines from collapsing.

closes: #1213

**NOTE:**
In the matrix chat room, it was mentioned that the expected behavior of `search.searchPage` was to open a new search tab even if one was already open. I assumed the intended behavior was to reuse the search tab. If you'd like this change to be reverted, let me know so I can add it to this PR

Screenshot:
Screen Shot 2020-03-11 at 14 47 18

### DIFF
--- a/extension/intents/search/queryScript.js
+++ b/extension/intents/search/queryScript.js
@@ -92,23 +92,52 @@ this.queryScript = (function() {
   }
 
   communicate.register("searchResultInfo", message => {
-    const cards = findCards();
-    const searchHeaders = document.querySelectorAll("a > h3");
+    const GOOGLE_SELECTOR = "a > h3";
+    const DDG_SELECTOR = "#links .results_links_deep .result__a";
+    const BING_SELECTOR = ".b_algo h2 a";
+    let searchEngine = "google";
+    let cards;
+
+    if (message && message.searchEngine && message.searchEngine.toLowerCase()) {
+      searchEngine = message.searchEngine.toLowerCase();
+    }
+
+    let selector;
+
+    switch (searchEngine) {
+      case "duckduckgo":
+        selector = DDG_SELECTOR;
+        break;
+      case "bing":
+        selector = BING_SELECTOR;
+        break;
+      default:
+        cards = findCards();
+        selector = GOOGLE_SELECTOR;
+    }
+
+    const searchHeaders = document.querySelectorAll(selector);
     const searchResults = [];
+
     for (const searchHeader of searchHeaders) {
       const parent = findParent(searchHeader, el => el.tagName === "LI");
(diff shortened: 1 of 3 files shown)


extension/intents/search/search.js
```diff
@@ -391,9 +391,25 @@ intentRunner.registerIntent({
...
-
       await content.lazyInject(tabId, "/intents/search/queryScript.js");
-      const searchInfo = await callScript({ type: "searchResultInfo" });
+
+      const engines = await browser.search.get();
+
+      const searchInfo = await callScript({
+        type: "searchResultInfo",
+        searchEngine: engines.find(engine => engine.isDefault).name,
+      });
```
ianb: I don't really know how expensive it is to get the search engine list each time, but I would worry about doing this. I think there's an easier test that can be done above

extension/intents/search/queryScript.js
```diff
@@ -92,23 +92,52 @@ this.queryScript = (function() {
...
-    const cards = findCards();
-    const searchHeaders = document.querySelectorAll("a > h3");
+    const GOOGLE_SELECTOR = "a > h3";
+    const DDG_SELECTOR = "#links .results_links_deep .result__a";
+    const BING_SELECTOR = ".b_algo h2 a";
+    let searchEngine = "google";
+    let cards;
+
+    if (message && message.searchEngine && message.searchEngine.toLowerCase()) {
+      searchEngine = message.searchEngine.toLowerCase();
```
ianb: I think it would be easier to simply use `location.origin` to test here. Google has other names, like `google.ca` etc, but testing for `/google/i.test(location.origin)` or `/duckduckgo/i.test(location.origin)` should be reliable enough.

JM-Mendez:
No worries about taking long, I'm sure you're swamped as it is. I'll get to work on your requested changes.

extension/intents/search/queryScript.js
JM-Mendez: Resolved using `location.origin` as you suggested.

extension/intents/search/search.js
JM-Mendez: Since you requested I use `location.origin` to retrieve the search engine being used, the `searchEngine` option to `callScript` became irrelevant and has been removed.

Issue #843, opened 2020-01-10
Add "find in this page" intent


ianb:
This would use the [find API](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/find/find) to search the current page.

I'm not sure exactly how it would work, so we should flush this out some more before implementation.

vibhuti019:
Please explain about the feature. I want to work on this.

awallin:
@vibhuti019 This feature would allow for a text voice search of the active web page. For example  saying something like:

* "find 'water' on this page"
* "search this page for 'water'"

Would open the find menu (also accessible via Command + F) and search for 'water'

ianb:
If someone wants to try this out: mostly you'll be exposing the find API that is mentioned in the description. So `find [query] on (this |) (page | tab)` or `search (this |) (page | tab) for [query]`

This will be a new intent, I guess in `tabs.js`. The _writing an intent_ document should give some guidance.

NoelyMarques:
Hi @ianb, I am an outreachy applicant and I'd like to have this issue assigned to me. May I?

shreyajain25:
Hi @ianb , May i work on this as my first issue?

vibhuti019:
I will have to work for mobile version only or for pc version too..

vibhuti019:
For mobile version we can give a mic icon so user can tap and perform their queries... 
As I could guess we can extend this too for other options too..

ianb:
I guess I will assign this to the first commenter, but @NoelyMarques please confirm in a comment here that you are working on it so that I can be sure it doesn't get stuck.

NoelyMarques:
Hello @ianb, yes I am!

Simpcyclassy:
I think this issue can be closed with the merged PR for verification

PR #1240, opened 2020-03-11
findOnPage intent #843

NoelyMarques:
This PR references  #843 

Added a new intent in tabs.toml called [tabs.findOnPage]
Registered new intent in tabs.js that find a query on page and makes it highlighted.
In case the word wouldn't be find a message will be showed.

### DIFF
--- a/extension/intents/tabs/tabs.js
+++ b/extension/intents/tabs/tabs.js
@@ -578,3 +578,20 @@ intentRunner.registerIntent({
     });
   },
 });
+
+intentRunner.registerIntent({
+  name: "tabs.findOnPage",
+  async run(context){
+   await browser.find.find(context.slots.query).then(found);
+  },
+});
+
+function found(results) {
+  if (results.count > 0) {
+    browser.find.highlightResults();
+  }else{
+    const e = new Error();
+    e.displayMessage = "Not found";
+    throw e;
+  }
+}
--- a/extension/intents/tabs/tabs.js
+++ b/extension/intents/tabs/tabs.js
@@ -582,16 +582,14 @@ intentRunner.registerIntent({
 intentRunner.registerIntent({
   name: "tabs.findOnPage",
   async run(context) {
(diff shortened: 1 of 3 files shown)


extension/intents/tabs/tabs.js
```diff
@@ -556,3 +556,20 @@ if (!buildSettings.android) {
     },
   });
 }
+
+intentRunner.registerIntent({
+  name: "tabs.findOnPage",
+  async run(context) {
+    await browser.find.find(context.slots.query).then(found);
```
ianb: You can use `const results = await browser.find.find(context.slots.query);`

Additionally I notice that if you find something that is scrolled off the page, then it will not actually scroll to that item. E.g., if I go to [this page](https://blog.rust-lang.org/2020/03/12/Rust-1.42.html) and say "find cargo on this page" it will highlight the term "cargo" but off-screen.

It looks like [browser.find.highlight](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/find/highlightResults) has an option to scroll. That could be called just after `browser.find.find`.

NoelyMarques: Hi @ianb!
I made the changes but I couldn't make the scroll to work. I used the [`{ noScroll: false }`](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/find/highlightResults) option . 
Is this the method you are referring to?

ianb: Yes, when I experimented I found `browser.find.highlight({noScroll: false, rangeIndex: 0})` worked (I guess it has to know which of the selections to scroll to)

NoelyMarques: That's it! Thanks :)

NoelyMarques:
Hi @ianb 
Could you have a look at it again?

ianb:
Cool, looks good, thanks!

PR #1244, opened 2020-03-12
Fixes #422 - Fix commands not working

### LINKED ISSUE #422: "Pause", "Stop", "Next", "Skip" commands not working for videos if Youtube tab was firstly opened in a channel page

melvin2016:

Fixes #422

Problem: 

- The dom has two elements of the same class names, which then were somehow colliding.
- The service was unable to differentiate between a channel page and a watch video page.


ianb:
Thanks! These weird arbitrary rules are where things get messy integrating with individual services, but I think this is the right way for now.

melvin2016:
@ianb Yes, since Youtube is a SPA I think it may be causing these errors.

Issue #1235, opened 2020-03-11
"Scroll to bottom" not working properly in a Youtube tab


alexandra-martin:
### Prerequisites:
Mic permissions are enabled.
"Allow" or "Don't Allow" button from “Allow Firefox Voice to Collect Voice Transcriptions” pop-up needs to be clicked.
"A privacy reminder from Google" is reviewed and agreed and the yellow language pop-up is closed. (if needed)

### STR:
1. Use the shortcut or click on the mic icon from the browser toolbar in a Youtube tab.
2. Say or write "Scroll to bottom", or "Scroll to end".

### Expected result:
The page is scrolled to the bottom of the page.

### Actual result:
Nothing happens.

### Notes:
Reproduced on Mac 10.14 and Win10x64 with Firefox Nightly 76.0a1 (2020-03-11).
If the Youtube page is not at the top, by making the "Scroll to end" command, the page goes up to the top.

![5](https://user-images.githubusercontent.com/54401091/76433818-8cc65b80-63bd-11ea-91bf-9fe381096973.gif)

JanviMahajan14:
hey @alexandra-martin  i would love to work on this issue.

naima-shk:
@alexandra-martin  Can i take this one please ?

saucekode:
@alexandra-martin I believe why did is an issue is because YouTube fetches more videos in real time as you scroll down with cases like comments and related videos, so as the command is given out, the height isn't defined yet due to more videos being fetched. But 'scroll down' works perfectly.

alexandra-martin:
@saucekode I was thinking the same thing. But if the page is scrolled in the middle for example, the "Scroll to end" makes the page go to the top, so maybe it's trying to do that also when it's already to the top, but it just can't get any higher. Don't know if it's somehow interfering with "Scroll to top", since these commands work fine on other pages.

saucekode:
I see your point. @ianb, can I work on this?

saucekode:
@alexandra-martin really, this is the developer playing safe. It's more like, since data is still being fetched, in this case, an undefined height, the right thing to do would be to scroll to top calculating the height from the middle which is defined. So scroll to end is synonymous to scroll to top in YouTube's case, I believe this would be the same for Instagram explore page too or even Facebook. Hopefully this makes sense @ianb do take a look at this, please

alexandra-martin:
[...]

PR #1245, opened 2020-03-12
Fixed Scrolling on pdf files

AbrilAnchondo:

Issue I was working on: https://github.com/mozilla/firefox-voice/issues/954
Found a solution by first checking if the file is a pdf, if so, the target element is the activeElement instead of the window.

### DIFF
--- a/extension/intents/scroll/scrollHelper.js
+++ b/extension/intents/scroll/scrollHelper.js
@@ -2,8 +2,13 @@
 
 const scrollAmount = 0.9;
 
+let url = location.href;
+let pdf = url.slice(-4) === ".pdf";
+let activeEl = document.activeElement;
+let targetEl = pdf ? activeEl : window;
+
 function scrollVertically(dy, smooth) {
-  window.scrollBy({
+  targetEl.scrollBy({
     left: 0,
     top: dy,
     behavior: smooth ? "smooth" : "auto",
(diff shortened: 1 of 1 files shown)


extension/intents/scroll/scrollHelper.js
```diff
@@ -2,8 +2,13 @@
 
 const scrollAmount = 0.9;
 
+const url = location.href;
+const pdf = url.slice(-4) === ".pdf";
```
ianb: This works for the one example, but generally it's not good to try to inspect a URL to understand the type of the document. @danielamormocea used `document.contentType`, which is a good way to check the content type... but that's actually not the right way to fix this anyway, as other sites like https://bugzilla.mozilla.org/show_bug.cgi?id=343205 or a Google Doc also aren't scrollable.

extension/intents/scroll/scrollHelper.js
```diff
@@ -2,8 +2,13 @@
...
+const activeEl = document.activeElement;
```
ianb: Note that `.activeElement` can change while a page is loaded (it follows focus), so this should be calculated each time.

ianb:
#1235 is also related to this

ianb:
I'm going to close this PR in favor of #1229

PR #1248, opened 2020-03-12
Dev/deezer

IkkyOdufade:
1. added deezer as a music service.
2. added an alert for non-registered users but having issues because i used window.alert().  Can you give me an alternative? I don't want it to throw an error. 


IkkyOdufade:
Waiting on your feedback @ianb. Thanks.

extension/services/deezer/deezer.js
```diff
@@ -0,0 +1,80 @@
+import * as music from "../../intents/music/music.js";
+import * as serviceList from "../../background/serviceList.js";
+import * as content from "../../background/content.js";
+import { shouldDisplayWarning } from "../../limiter.js";
+
+class Deezer extends serviceList.Service {
```
ianb: You should be able to reduce this code if you use `import MusicService from "../../background/musicService.js";` like `soundcloud.js`

extension/services/deezer/deezer.js
```diff
@@ -0,0 +1,80 @@
...
+  async playQuery(query) {
+    await this.initTab("/services/deezer/player.js");
+    if(this.initTab){
+   window.alert ("Please signup or login to use this service. If you have, please ignore this message.");
```
ianb: I think you can do:

```js
this.context.activateTab(this.tab);
const exc = new Error("Not logged in");
exc.displayMessage = "Please log in";
throw exc;
```

That said, this probably isn't the right place to raise this exception. The tab might get opened but the user might still be logged in. Instead you should let `player.js` raise the exception and catch it

extension/services/deezer/deezer.js
```diff
@@ -0,0 +1,80 @@
...
+    }
+    await this.callTab("search", { query, thenPlay: true });
```
ianb: So here there should be something like:

```
try {
  await this.callTab(...);
} catch (e) {
  if (e.message === "Not logged in") {
    e.displayMessage = "Please log in";
    this.context.activateTab(this.tab);
  }
  throw e;
}
```

And maybe this should go in `musicPlayer.js`

extension/services/deezer/player.js
```diff
@@ -0,0 +1,52 @@
+/* globals helpers */
+
+this.player = (function() {
+  const SEARCH_PLAY = ".datagrid-row song a[role='button']";
+
+  class Player extends helpers.Runner {
+    action_play() {
+      const button = this.querySelector("button[aria-label='Play']");
```
ianb: I realize spotify works like this currently, but if it's possible to not use the labels that would be better: then if someone is using a localized version of the site, these selectors will still work.

README.md
```diff
@@ -91,24 +91,6 @@ You will need to setup Firefox Nightly or Developer on WSL before running `firef
...
-Before cloning the repo, using the terminal in Git-bash:
-
-1. Run `git config --global core.autocrlf false` to prevent git from converting line endings from LF to CRLF
-2. Run `npm config set script-shell "path\\to\\bash.exe"`. Example of path is `"C:\\Program Files\\git\\bin\\bash.exe"`. This enables npm to run linux-like commands.
-3. Clone the [repository](https://github.com/mozilla/firefox-voice.git) `git clone https://github.com/mozilla/firefox-voice.git`
-4. To create the environment variable `FIREFOX`, run `export FIREFOX=normalized/path/to/firefox.exe`. For example if the install path is `C:\Program Files\Firefox Nightly\firefox.exe` then normalized path is `/c/Program Files/Firefox Nightly/firefox.exe`. Alternatively, the environment variable set to the normalized path can be created using the Windows system dialog.
-5. Now run `npm install`
-6. Run `npm start`. This will launch a new Firefox browser with the `firefox-voice` extension installed.
-
-
```
ianb: I'm guessing you didn't mean to include this change in the PR...?

IkkyOdufade:
Ok. Thank you.

README.md
IkkyOdufade: I didnt....

IkkyOdufade:
Hi @ianb . Please go through my corrections and I'll be waiting for your feedback. Thank you.

README.md
ianb: I think if you run `git checkout origin/master -- README.md` it should undo this.

extension/background/MusicService.js
```diff
@@ -0,0 +1,69 @@
+import * as serviceList from "../../background/serviceList.js";
```
ianb: This should be using the existing file in `extension/background/musicService.js` (if you don't see that file you'll need to merge master into your branch/fork)

IkkyOdufade: Hi @ianb . Thank you for your corrections. Would work on it now.

IkkyOdufade:
Hi Ian. I had to close this pr because I could not get git to do what I wanted. So I created a new branch and opened a new pr, with just deezer. Thank you.

PR #1161, opened 2020-03-06
Fix Issue Make format linting part of tests #1144

lelouchB:
Fixed Issue #1144

PR #1249, opened 2020-03-12
Fix Issue Make format linting part of tests #1144 

### LINKED ISSUE #1144: Make format linting part of tests

lelouchB:
@ianb I have closed PR #1161  and opened this one with less commits and same file changes
Fixes #1144 


ianb:
Sorry this sat for so long, I had to get around to the process of checking and reverting a couple PRs. Looks good, has a failure... but there are some other failures in master anyway that will have to be fixed with this linting applied.

PR #1263, opened 2020-03-12
Handles exception if no search result found

### LINKED ISSUE #702: "Doorhanger" not displayed properly if a long unbroken string command is made
### Prerequisites:
Mic permission are enabled.
"A privacy reminder from Google" is reviewed and agreed and the yellow language pop-up is closed.

### STR:
2. Write an unbroken string of 129 characters and press the "Enter" key.

### Expected result:
There are no display issues.

### Actual result:
There are display issues in how the "Doorhanger" is shown.

### Notes:

GangaChatrvedi:
Fixes #702 

### DIFF
--- a/extension/intents/search/search.js
+++ b/extension/intents/search/search.js
@@ -304,6 +304,13 @@ intentRunner.registerIntent({
     await performSearch(context.slots.query);
     const searchInfo = await callScript({ type: "searchResultInfo" });
     searchInfo.query = context.slots.query;
+
+    if (!searchInfo.searchResults) {      
+      const e = new Error("No result found for "+searchInfo.query);
+      e.displayMessage = "No result found for "+searchInfo.query;
+      throw e;
+    }
+
     if (searchInfo.hasCard || searchInfo.hasSidebarCard) {
       const card = await callScript({ type: "cardImage" });
       context.keepPopup();
--- a/extension/intents/search/search.js
+++ b/extension/intents/search/search.js
@@ -305,9 +305,9 @@ intentRunner.registerIntent({
     const searchInfo = await callScript({ type: "searchResultInfo" });
     searchInfo.query = context.slots.query;
 
-    if (!searchInfo.searchResults) {      
-      const e = new Error("No result found for "+searchInfo.query);
-      e.displayMessage = "No result found for "+searchInfo.query;
+    if (!searchInfo.searchResults) {
+      const e = new Error("No result found for " + searchInfo.query);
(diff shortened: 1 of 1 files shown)


ianb:
@GangaChatrvedi #702 is about long strings in the doorhanger. I'm a little unclear what this is fixing? Can you give some details on what you have to do to get the bug being fixed here?

GangaChatrvedi:
@ianb I think layout issue in #702 is because of an exception when there is no search result for query so this is fixing that issue by handling exception.

ianb:
@GangaChatrvedi #702 only looks like a layout issue, something like `text-overflow` (or other CSS) is probably all that's needed to fix it. The search result shouldn't really matter.

It's possible you encountered a regression from #1180 which he's trying to fix in #1236 ...?

GangaChatrvedi:
@ianb I was taking a look again at #702 and reconfirmed that it is certainly not a css issue as with 129 char string we have a complete blank screen on doorhanger. When checked it leaded to exception of no result found.

This all is happening because of search rule of google
- max 50 words
- each word of max 128 char

We have one of 2 possible solution to overcome this error.
- Trim each word by 128 char
- Show that we could not find any result. Because its came thing that google is going as well on general search page.


This will work
https://www.google.com/search?q=hhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhh

This will not
https://www.google.com/search?q=hhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhh

ianb:
While trying to understand this I keep getting [this fine image](https://imgur.com/t/hhhhhhhhhhhhhhhhhhhhhhhhh/oW26UT4). Thanks internet!

Anyway, I guess this is OK, or at least doesn't effect anything badly, but I haven't really been able to get to any error condition myself.

alexandra-martin:
> I think layout issue in #702 is because of an exception when there is no search result for query

Good call @GangaChatrvedi. Indeed, with even a smaller string, that doesn't give search results on Google, the same behavior appears.

PR #1279, opened 2020-03-13
Fixes #1109 Turn Selection into link 

### LINKED ISSUE #1109: Intent: turn selection into link
This would take the current selection and turn it into a link. If the selection is in a contenteditable then it will be a proper HTML link, otherwise a Markdown link.

The link would be calculated first by looking at the other open tabs, and then if nothing matched it would be a Google I'm Lucky search.

JanviMahajan14:
 #1109
 
@ianb  i have tried to solve this but not sure is this a right approach to solve this.

### DIFF
--- a/extension/intents/forms/forms.js
+++ b/extension/intents/forms/forms.js
@@ -1,18 +1,6 @@
-/* globals buildSettings */
-
 import * as intentRunner from "../../background/intentRunner.js";
 import * as content from "../../background/content.js";
 import * as pageMetadata from "../../background/pageMetadata.js";
-import * as searching from "../../searching.js";
-
-function createTabGoogleLucky(query, options = {}) {
-  const searchUrl = searching.googleSearchUrl(query, true);
-  const tab = browser.tabs.create({ url: searchUrl });
-  if (options.hide && !buildSettings.android) {
-    browser.tabs.hide(tab.id);
-  }
-  return searchUrl;
-}
 
 intentRunner.registerIntent({
   name: "forms.dictate",
@@ -96,7 +84,12 @@ intentRunner.registerIntent({
       e.displayMessage = "No text selected";
       throw e;
     }
-    const url = createTabGoogleLucky(selection.text);
+    const newTab = await context.createTabGoogleLucky(selection.text, {
+      hide: true,
+    });
+    const url = newTab.url;
+    const title = newTab.title;
+    await browser.tabs.remove(newTab.id);
     await content.lazyInject(activeTab.id, [
       "/js/vendor/fuse.js",
       "/intents/forms/formsContentScript.js",
@@ -104,6 +97,7 @@ intentRunner.registerIntent({
     await browser.tabs.sendMessage(activeTab.id, {
       type: "turnSelectionIntoLink",
       url,
+      title,
     });
   },
 });
--- a/extension/intents/forms/forms.js
+++ b/extension/intents/forms/forms.js
@@ -88,7 +88,7 @@ intentRunner.registerIntent({
       hide: true,
     });
     const url = newTab.url;
-    const title = newTab.title;
+    const text = selection.text;
     await browser.tabs.remove(newTab.id);
     await content.lazyInject(activeTab.id, [
       "/js/vendor/fuse.js",
@@ -97,7 +97,7 @@ intentRunner.registerIntent({
     await browser.tabs.sendMessage(activeTab.id, {
       type: "turnSelectionIntoLink",
       url,
-      title,
+      text,
     });
   },
 });
--- a/extension/intents/forms/forms.js
+++ b/extension/intents/forms/forms.js
@@ -7,7 +7,7 @@ import * as searching from "../../searching.js";
 
 function createTabGoogleLucky(query, options = {}) {
   const searchUrl = searching.googleSearchUrl(query, true);
-  const tab =  browser.tabs.create({ url: searchUrl });
+  const tab = browser.tabs.create({ url: searchUrl });
   if (options.hide && !buildSettings.android) {
     browser.tabs.hide(tab.id);
   }
@@ -102,8 +102,8 @@ intentRunner.registerIntent({
       "/intents/forms/formsContentScript.js",
     ]);
     await browser.tabs.sendMessage(activeTab.id, {
-       type: "turnSelectionIntoLink",
-       url,
+      type: "turnSelectionIntoLink",
+      url,
     });
   },
 });
--- a/extension/intents/forms/forms.js
+++ b/extension/intents/forms/forms.js
@@ -1,5 +1,18 @@
+/* globals buildSettings */
+
 import * as intentRunner from "../../background/intentRunner.js";
 import * as content from "../../background/content.js";
+import * as pageMetadata from "../../background/pageMetadata.js";
+import * as searching from "../../searching.js";
+
+function createTabGoogleLucky(query, options = {}) {
+  const searchUrl = searching.googleSearchUrl(query, true);
+  const tab =  browser.tabs.create({ url: searchUrl });
+  if (options.hide && !buildSettings.android) {
+    browser.tabs.hide(tab.id);
+  }
+  return searchUrl;
+}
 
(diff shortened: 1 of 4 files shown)


extension/intents/forms/forms.js
```diff
@@ -1,5 +1,18 @@
+/* globals buildSettings */
+
 import * as intentRunner from "../../background/intentRunner.js";
 import * as content from "../../background/content.js";
+import * as pageMetadata from "../../background/pageMetadata.js";
+import * as searching from "../../searching.js";
+
+function createTabGoogleLucky(query, options = {}) {
```
ianb: You should add the hide option to the `createTabGoogleLuck` method in `intentRunner.js`. That function does some other important things you'll want to use for this as well.

extension/intents/forms/forms.js
```diff
@@ -72,3 +85,25 @@ intentRunner.registerIntent({
...
+  name: "forms.turnSelectionIntoLink",
+  async run(context) {
+    const activeTab = await context.activeTab();
+    const selection = await pageMetadata.getSelection(activeTab.id);
+    if (!selection || !selection.text) {
+      const e = new Error("No text selected");
+      e.displayMessage = "No text selected";
+      throw e;
+    }
+    const url = createTabGoogleLucky(selection.text);
```
ianb: Whoops, looks like you didn't actually hide the tab either.

extension/intents/forms/forms.js
```diff
@@ -72,3 +85,25 @@ intentRunner.registerIntent({
...
+    await content.lazyInject(activeTab.id, [
+      "/js/vendor/fuse.js",
+      "/intents/forms/formsContentScript.js",
+    ]);
+    await browser.tabs.sendMessage(activeTab.id, {
+      type: "turnSelectionIntoLink",
+      url,
+    });
```
ianb: Once everything is done you can close the tab you opened above.

extension/intents/forms/formsContentScript.js
```diff
@@ -18,6 +18,7 @@ this.dictationContentScript = (function() {
   ];
   const COMBINED_SELECTORS = PASTE_SELECTORS.join(", ");
   const DEFAULT_FOCUS_TIME = 500;
+  const meta = pageMetadataContentScript.getMetadata;
```
ianb: I don't think you should need this

extension/intents/forms/formsContentScript.js
```diff
@@ -192,6 +193,28 @@ this.dictationContentScript = (function() {
...
+          break;
+        }
+      }
+    } else {
+      el = document.activeElement;
+    }
+    if (el.hasAttribute("contenteditable")) {
+      replaceSelectedText(url);
+    } else if (el.hasAttribute("textarea")) {
+      const m = meta();
```
ianb: This is running in the page where the link is being created, so it's going to pick up the title and URL of the current page. You want to replace it with the title and URL that you found for the query.

extension/intents/forms/formsContentScript.js
```diff
@@ -233,4 +256,19 @@ this.dictationContentScript = (function() {
...
   }
+
+  function replaceSelectedText(replacementText) {
+    let sel, range;
+    if (window.getSelection) {
+      sel = window.getSelection();
+      if (sel.rangeCount) {
+        range = sel.getRangeAt(0);
+        range.deleteContents();
+        range.insertNode(document.createTextNode(replacementText));
```
ianb: This should be doing something like:

```js
const anchor = document.createElement("a");
anchor.textContent = title;
anchor.href = url;
range.insertNode(anchor);
```

Then it will insert a new rich link.

extension/intents/forms/forms.js
JanviMahajan14: @ianb  sorry as i have misunderstood the things as horribly wrong. After updating the changes suggested by you I am still getting some problems. Like here the `newTab.title` is not giving an appropriate tiltle and seems to give something like url

JanviMahajan14:
@ianb  sorry as i have misunderstood the things as horribly wrong. After updating the changes suggested by you I am still getting some problems. Like in the `newTab.title` is not giving an appropriate tiltle and seems to give something like url. And also in formContentScript.js  in the `if (el.hasAttribute("contenteditable"))` the code within the if else block is not working . Do i need to take something else for`el`

extension/intents/forms/forms.js
```diff
@@ -96,14 +84,20 @@ intentRunner.registerIntent({
       e.displayMessage = "No text selected";
       throw e;
     }
-    const url = createTabGoogleLucky(selection.text);
+    const newTab = await context.createTabGoogleLucky(selection.text, {
+      hide: true,
+    });
+    const url = newTab.url;
+    const title = newTab.title;
```
ianb: It just occurred to me that this should actually keep `selection.text` as the text of the link.

extension/intents/forms/formsContentScript.js
```diff
@@ -257,18 +256,21 @@ this.dictationContentScript = (function() {
...
+  function replaceSelectedText(title, url) {
     let sel, range;
     if (window.getSelection) {
       sel = window.getSelection();
       if (sel.rangeCount) {
         range = sel.getRangeAt(0);
         range.deleteContents();
-        range.insertNode(document.createTextNode(replacementText));
+        const anchor = document.createElement("a");
+        anchor.textContent = title;
```
ianb: I.e., `anchor.textContent` should be the selection text.

extension/intents/forms/forms.js
ianb: This should still close. I.e., call `browser.tabs.remove(newTab.id)`

JanviMahajan14:
@ianb  I have called `browser.tabs.remove(newTab.id)` just after fetching the URL and title from the new tab. Can you tell me if I should make some changes in the `el.hasAttribute()` as the if else block is not working do I need to take something else as `el`

extension/intents/forms/formsContentScript.js
```diff
@@ -192,6 +192,28 @@ this.dictationContentScript = (function() {
...
+      for (const selector of PASTE_SELECTORS) {
+        el = document.querySelector(selector);
+        if (el && isInViewport(el)) {
+          break;
+        }
+      }
+    } else {
+      el = document.activeElement;
+    }
+    if (el.hasAttribute("contenteditable")) {
```
ianb: Oh, I see the issue here. So with contenteditable you might have something like:

```html
<div contenteditable="1">
  <p>
    Here is some text I can edit.
  </p>
</div>
```

And the selection will be inside the `<p>`. So you have to look at the parent, like:

```js
let isContentEditable = false;
let checkElement = el;
while (checkElement) {
  if (checkElement.hasAttribute("contenteditable")) {
    isContentEditable = true;
    break;
  }
  checkElement = checkElement.parentNode;
}
```

JanviMahajan14:
@ianb  updates changes!

extension/intents/forms/formsContentScript.js
```diff
@@ -206,11 +206,16 @@ this.dictationContentScript = (function() {
...
-    if (el.hasAttribute("contenteditable")) {
-      replaceSelectedText(text, url);
-    } else if (el.hasAttribute("textarea")) {
-      const markdownLink = `[${text}](${url})`;
-      replaceSelectedText(text, markdownLink);
+    while (el) {
+      if (el.hasAttribute("contenteditable")) {
+        replaceSelectedText(text, url);
+        break;
+      } else if (el.hasAttribute("textarea")) {
```
ianb: Textareas are an element, so a test like `el.tagName === "TEXTAREA"` would be correct here

JanviMahajan14:
@ianb updated changes for textarea

ianb:
`"TEXTAREA"` has to be upper-case (`.tagName` just happens to return upper case). But there's too many in-flight PRs so instead of commenting/revising/rereviewing I'm just going to land it and fix it. Onward and upward!

I think this is a cool feature, our next big challenge will be figuring out how to tell people about features like this.

Issue #1267, opened 2020-03-12
[Intent] Count my open tabs


awallin:
Based on telemetry we've seen people ask for the total number of their tabs by saying things like “count my open tabs” or “count how many tabs I have open” or “count my tabs”. I lean towards this being the total number of tabs on the current window but could be convinced it's total number of tabs on all windows?

The result should be displayed in the doorhanger matching attached mockups. Details to note:
* Font size for resulting number 92
* Education text is: Click mic and say ‘gather all Google tabs’

![open-tabs](https://user-images.githubusercontent.com/428245/76588866-71cf1680-64a5-11ea-9b5a-cdffcc509e06.jpg)

xlisachan:
@awallin can I work on this?

awallin:
Go for it @xlisachan

awallin:
Made a few changes to the content and visual layout for this to establish a template for Q&A cards used in #1266 #1270 and #1269. Specs for these cards are as follows:

* 100x100 image with
* 25px of padding between Question Transcript and image
* 35px of padding between image and Answer text

donalawa:
Can i also work on this?

awallin:
@donalawa This is a straight forward issue. Keep an eye out for more 'good first issues' coming soon.

xlisachan:
@awallin Please take a look at the updates and notes I included in the pull request mentioned above when you have a moment. Thank you!

alexandra-martin:
Verified on Mac 10.14 and Win 10x64 with Firefox Nightly 76.0a1 (2020-03-18).

![1](https://user-images.githubusercontent.com/54401091/76960090-71aa9d00-6923-11ea-9a77-834314ed56a8.gif)

PR #1281, opened 2020-03-13
RE: [Intent] Count my open tabs 1267

xlisachan:
Based on the intents mentioned in #1267, I think it also makes sense to provide the number of tabs on the current window only, as opposed to the total open tabs of all windows. Please see a screenshot of the change below.

Screen Shot 2020-03-13 at 4 35 48 PM

I noticed that after making a query like "Who created Breaking Bad?", it counts the hidden pinned search tab (indicated on line 42 of search.toml), which may confuse users. The Telemetry ping in the console also counts the hidden tab as well. Can you please review and provide guidance when you have a moment?

Screen Shot 2020-03-13 at 5 05 10 PM

Screen Shot 2020-03-13 at 5 03 54 PM

### DIFF
--- a/extension/intents/tabs/tabs.js
+++ b/extension/intents/tabs/tabs.js
@@ -562,9 +562,11 @@ intentRunner.registerIntent({
   async run(context) {
     context.keepPopup();
     const tabs = await browser.tabs.query({ currentWindow: true });
+    const hiddenTabs = await browser.tabs.query({ hidden: true });
+    const numOfOpenTabs = tabs.length - hiddenTabs.length;
     const card = {
       answer: {
-        num: `${tabs.length}`,
+        num: `${numOfOpenTabs}`,
         text: "Open tabs",
         eduText: "Click mic and say 'gather all Google tabs'",
       },
--- a/extension/intents/tabs/tabs.js
+++ b/extension/intents/tabs/tabs.js
@@ -556,3 +556,23 @@ if (!buildSettings.android) {
     },
   });
 }
+
+intentRunner.registerIntent({
+  name: "tabs.countTabs",
+  async run(context) {
+    context.keepPopup();
+    const tabs = await browser.tabs.query({ currentWindow: true });
+    const card = {
+      answer: {
+        num: `${tabs.length}`,
+        text: "Open tabs",
+        eduText: "Click mic and say 'gather all Google tabs'"
+      }
+    };
+    await browser.runtime.sendMessage({
+      type: "showSearchResults",
+      card,
+      searchResults: card,
+    });
+  },
+});
--- a/extension/intents/tabs/tabs.js
+++ b/extension/intents/tabs/tabs.js
@@ -566,8 +566,8 @@ intentRunner.registerIntent({
       answer: {
         num: `${tabs.length}`,
         text: "Open tabs",
-        eduText: "Click mic and say 'gather all Google tabs'"
-      }
+        eduText: "Click mic and say 'gather all Google tabs'",
+      },
     };
     await browser.runtime.sendMessage({
       type: "showSearchResults",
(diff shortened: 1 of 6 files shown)


ianb:
With regard to hidden tabs, I think you are right that both here and in telemetry.js we should exclude hidden tabs from the count. It looks like `.query({hidden: false})` will do that.

extension/intents/tabs/tabs.js
```diff
@@ -556,3 +556,23 @@ if (!buildSettings.android) {
...
+intentRunner.registerIntent({
+  name: "tabs.countTabs",
+  async run(context) {
+    context.keepPopup();
+    const tabs = await browser.tabs.query({ currentWindow: true });
+    const card = {
+      answer: {
+        num: `${tabs.length}`,
+        text: "Open tabs",
+        eduText: "Click mic and say 'gather all Google tabs'",
```
ianb: It looks like we actually don't support "gather", though that could be added to the "collect" phrases that we have (`(collect | gather)`).

I noticed @awallin put single quotes around `'gather all Google tabs'` in the mockup, but generally we've been using double quotes (`"gather all Google tabs"`).

Also could you rename `num` to be `largeText`? It doesn't have to be a number, but it will be very large! I can imagine `mediumText` being added soon as well for some other information.

extension/intents/tabs/tabs.js
```diff
@@ -556,3 +556,23 @@ if (!buildSettings.android) {
...
+      },
+    };
+    await browser.runtime.sendMessage({
+      type: "showSearchResults",
```
ianb: Hmm... we should probably make a followup to rename this, I can see why it would be the easy thing to work with here, but "showSearchResults" also isn't the right name anymore.

extension/popup/popupView.jsx
```diff
@@ -589,23 +595,50 @@ const SearchResultsContent = ({
     onNextSearchResultClick();
   };
 
+  const searchCard = () => (
```
ianb: We've been capitalizing these functions that return markup, so SearchCard would be a better name.

extension/popup/popupView.jsx
```diff
@@ -589,23 +595,50 @@ const SearchResultsContent = ({
...
+    <button class="invisible-button" onClick={onSearchCardClick}>
+      <img id="search-image" alt={imgAlt} style={cardStyles} src={card.src} />
+    </button>
+  );
+
+  const qaCard = () => (
```
ianb: I don't understand the name `qaCard`? Maybe `AnswerCard` would be the better name?

xlisachan: It was referred to as the Q&A card in issue #1267 so I named it `qaCard`, haha. `AnswerCard` is definitely a better name, I will make this change.

extension/popup/popupView.jsx
xlisachan: I will make this update!

extension/intents/tabs/tabs.js
xlisachan: I make updates to support "gather", use double quotes, and rename `num`.

extension/intents/tabs/tabs.js
xlisachan: I was actually in the middle of creating a new one for (what is now) AnswerCard, but I saw a note saying this issue should be a straightforward update, so I thought I was going deep down a rabbithole. If it's decided that it will be renamed, I would like to give it another shot!

xlisachan:
@ianb Thanks for all the feedback! I have made the following updates:
1. Excluded hidden tabs in the counter in tabs.js
2. Excluded hidden tabs in the numberOfTabs counter in telmetry
3. Added support for "gather" in tabs.toml
4. Updated single quotes to double quotes in IntentFeedback tag
5. Changed variable `num` to `largeText` and updated the className as well
6. Capitalized markup function `SearchCard()` 
7. Renamed `qaCard` to `AnswerCard()`

Please review and let me know if you have any questions.

ianb:
Looking really good, thanks! And you got some extra improvements in there along the way...

PR #1288, opened 2020-03-15
Fixes #68

### LINKED ISSUE #68: Allow other extensions to add intent handlers
We'd like other extensions to be able to extend the capabilities of this project.

An open issue: how do we extend the intent parser given these extensions?

Two options for extensions that could support Firefox Voice to demonstrate how this works:
* [Email Tabs](https://github.com/mozilla/email-tabs): composes emails based on one or more tabs
* [Side View](https://github.com/mozilla/side-view): opens a mobile view of a tab in the sidebar

(These are good options because I developed them and can make the changes.)

lelouchB:
FIxes #68 @ianb  Here is the draft PR
Sending and Receiving Message is working fine
I am also attaching a screenshot
inet
 

ianb:
@lelouchB Yes, this looks like the right direction to go. Then you'll want to do something like make the bordify extension add an intent like "add borders to this page" that will cause the borderify extension to run.

lelouchB:
You mean saying into Firefox voice runs borderify?

ianb:
Firefox Voice should send a message to borderify, telling it to run the intent. Basically what I outlined in the comment: https://github.com/mozilla/firefox-voice/issues/68#issuecomment-598804363

lelouchB:
@ianb Understood, Will get on it right away

lelouchB:
@ianb Done, Made an intent which triggers run of some commands in Borderify
What should be next step?

lelouchB:
@ianb Do you have any advice for the next step?

ianb:
@lelouchB I'm sorry, with the PR title I'm afraid I kept skimming over this one.

Ultimately there should be no borderify-specific changes in `extension/` – instead this should add hooks that allow borderify to tell Firefox Voice about itself.

So Firefox Voice will need something like:

```js
browser.runtime.onMessageExternal((message, sender) => {
  if (message.type === "registerIntent") {
    for (const intent of message.intents) {
      registerExternalIntent(intent, sender);
    }
  }
});

function registerExternalIntent(intent, sender) {
  intent.extensionId = sender.id;
  intent.run = runExternalIntent;
  intentRunner.registerIntent(intent);
}

async function runExternalIntent(context) {
  const resp = await browser.runtime.sendMessage(this.extensionId, {
    type: "runIntent",
    context,
  });
  // Any methods or property changes on context will have to come back through resp
}
```

There will also be several changes required elsewhere to make this all work, but this is roughly how it would start.

lelouchB:
Cool, thanks @ianb 
Be safe

Issue #995, opened 2020-02-14
Intent to make videos fullscreen


ianb:
#913 (PR #985) added a full screen intent, but sometimes what's intended is to make the _video_ fullscreen, not the browser/page.

I think this should look at the active tab, and probably use the services infrastructure (e.g., `extension/services/youtube/youtube.js`). I'm not sure if there's actually a general way to make a video full screen.

xlisachan:
@ianb I tried to solve this issue, but keep getting the same error (details are included in the pull request above). If you could please provide some guidance during the week (apologies for commenting on the weekend!!), it would be greatly appreciated!

PR #1289, opened 2020-03-15
Attempt to add intent to make video full screen

xlisachan:
I was looking into this issue #995 and could use your expertise when you have a moment, as I ultimately keep getting the error below.

Screen Shot 2020-03-15 at 4 11 38 PM

As advised in the issue, I updated the current fullscreen intent, so that new phrases were not needed - though I did move the intents from `extension/intents/tabs/tabs.js` to `extension/intents/music/music.js` so I would have access to `extension/services/youtube/player.js`.

Attempts:
1. Utilized `.requestFullscreen()`; got the error above -- `mozRequestFullScreen()` has been deprecated
2. Tried to target the button instead as the other controls were solved; still got the error above
3. In devtools, I targeted the button in the console; still got the error. (Sidenote: the button does not have an aria-label or title like the other buttons).

I'm not sure if the response time isn't short enough (1 sec), or it's just not possible? Any guidance would be greatly appreciated!

### DIFF
--- a/extension/intents/music/music.js
+++ b/extension/intents/music/music.js
@@ -130,7 +130,10 @@ intentRunner.registerIntent({
       exc.displayMessage = exc.message;
       throw exc;
     }
-    const audibleTab = await browser.tabs.query({ active: true, audible: true });
+    const audibleTab = await browser.tabs.query({
+      active: true,
+      audible: true,
+    });
     if (!audibleTab.length) {
       const currentWindow = await browser.windows.getCurrent();
       await browser.windows.update(currentWindow.id, { state: "fullscreen" });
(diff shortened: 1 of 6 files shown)


ianb:
@xlisachan Hmmm... I'm afraid this is going to turn out to be quite hard. Many permissions like this can only be granted in response to a "user initiated event". In other words the user has to actually click a button or type a key, and if they don't then Firefox will reject the attempt without any prompting (just the message in the console that you are seeing). Doing `.click()` – or just about anything else – will not work (it uses the `.isTrusted` property on events: https://developer.mozilla.org/en-US/docs/Web/API/Event)

Fixing this would be quite challenging, as we'd have to get in Firefox's internals to handle this. It's the same basic issue that also keeps us from autoplaying when we open a tab, so we probably should fix it someday, but it will be hard.

xlisachan:
@ianb Makes sense. I'll go ahead and close this pull request then, thanks for the update!

Issue #841, opened 2020-01-09
[Feature] Firefox Voice should remember the context / search current page


firefoxvoiceissuebot:
## Feature Request
Firefox Voice should remember the context of the statement.

I said "Open Youtube" and then said "Search for CutCapers" (a band), but what ended up happening was Youtube opening in one tab, and Google opening in another, with a search for "cut capers".

When I combined the statements by saying "Search Youtube for CutCapers" it worked... fine enough (It opened the first video that Youtube finds when you search "CutCapers").

Additionally Google isn't my default search engine, but I can deal with that.

ianb:
I think we could add a new intent, "search this site for [query]" (or page, or "search here for [query]" and some other similar patterns).

In that case, we would try to look up the domain in the list of bang searches (see navigation.js), find the right one, and then do a search in the current tab.

farhatcode:
Hi ian, Can I work on this issue? I am Outreachy applicant.

danielamormocea:
I will start working on this, if that's alright :) (also from Outreachy)

PrastutiUpadhaya:
Hi, can I please work on this?

ianb:
@farhatcode Sure, give it a go. Be aware this will involve several changes. I realize we don't actually have a mapping of domain names to bang searches, so for instance we don't know that "amazon", which maps to the `az` bangSearch, is also related to `amazon.com`. You might need to add domain names to `searchServices.toml` and then use that to check if the domain name of the active tab (`(new URL((await browserUtils.activeTab()).url)).hostname`) matches up with one of those services.

It might be good to try to handle just one of the services and make a draft PR to get feedback.

farhatcode:
Ok Ian, I am starting to work on this issue.

farhatcode:
Referring to # 1194, How we can differentiate that the user wants to search in the currently open site or he wants to search in the default search provider, when he first says "open youtube" and then says "Search for X" and he intends to search youtube and want to see the results in the active tab.

ianb:
@farhatcode I think we should make it explicit, so the user will have to say "search this site for [query]" or "search for [query] on this tab" or something specific. So not exactly like described in the original ticket, it won't override a normal search phrase like "search for cutcapers"

PR #1291, opened 2020-03-16
Draft - Remember search context

farhatcode:
**Edits registerIntent() for navigation.bangSearch in navigation.js**

Considers 3 cases for navigation.bangSearch :
**1.** supported service + serviceName mentioned 
e.g. "search amazon for X"
behavior: opens amazon search results in a new tab 
**2.** supported service + reference of active tab | site used (serviceName not mentioned)
e.g. "search here | this tab for Y"
behavior: searches using the active tab's service and opens search results in the same tab. 
**3.** unsupported service
e.g. open bing.com, say "search for news here"
behavior: uses default search engine and brings search results for news from bing.com
(Case 1 was already implemented, a little bit of modification done to add other 2 cases. I am adding case 2 and 3 to code.)

Adding match patterns and examples to navigation.toml

Adding baseUrl for amazon and coursera only, in searchServices.toml, for draft PR.

Draft PR for #841


farhatcode:
After review / changes / approval, I will add the same behavior for music services and if required for search.searchPage too.

farhatcode:
@ianb I have my draft PR ready for issue #841. Please give your feedback when you find some time.

farhatcode:
@ianb this is another draft PR awaiting your feedback.

extension/intents/navigation/navigation.js
```diff
@@ -66,15 +67,37 @@ intentRunner.registerIntent({
...
+      myurl = new URL((await browserUtil.activeTab()).url);
+      const tab = await context.activeTab();
+      await browser.search.search({
+        query:
+          myurl.origin !== "null" && myurl.href !== "about:blank"
+            ? context.slots.query + ` site:${myurl.origin}`
+            : context.slots.query,
+        tabId: tab.id,
+      });
+    }
```
farhatcode, line 99: This else part displays search results in default search engine with search bar displaying:
“[query] site:https://www.example.com”
For example, I opened yahoo.com and gave command "search sports here", above code gives results like this: 
https://duckduckgo.com/?t=ffsb&q=sports+site%3Ahttps%3A%2F%2Fwww.yahoo.com&ia=web

Instead of displaying the search results from the default search engine, I want to perform  search using the relevant service and display results from that site. To serve the purpose of this issue (which was: **search here**)
I want to change this else part to something like this: 
```
const tabOrigin = new URL((await browserUtil.activeTab()).url).origin;
const startIndex = tabOrigin.indexOf(".");  
const endIndex = tabOrigin.indexOf(".", startIndex+1);  
const unKnownService = tabOrigin.substring(startIndex+1, endIndex); 
myurl = await searching.searchUrlUnknownService(context.slots.query, unKnownService); 
await browser.tabs.update({ url: myurl });
```
Keeping searchUrlUnknownService() method similar to ddgBangSearchUrl(), except it does not check metadata for service bang, it is given the service name from the active tab’s url and it generates the search url with given service name.
Repeating the same example:  I opened yahoo.com and gave command "search sports here". This code gives results which are more desirable, like this: 
https://search.yahoo.com/search?p=sports

farhatcode: Now what I want to do is to search https://duckduckgo.com/bang to bring in the bang from ddg.
I recently noticed from #1475, that we can do searches in a hidden tab and use the results. 
I am able to redirect to ddgbang searches to get the right bang. I believe we can bring bangs from ddg and redirect to any website the user wants. 
@ianb can you guide me in the right direction and code, How can I perform any task in hidden tab and bring in the search results from there?

Issue #76, opened 2019-08-13
Native timer


ianb:
Timer intents would support:

* Start timer
* Start timer for [X]
* Pause timer
* Stop timer
* Change timer to [X]

We could implement a timer natively, or attempt some integration. Technically Google supports timers [via search](https://www.google.com/search?client=firefox-b-1-d&q=set+timer+for+10+minutes), but we'd have to leave the page open, and the interface isn't particularly nice.

awallin:
Adding details and mockups to implement a native timer.

* Timer can be set with phrases such as "set a timer for 5 minutes", "10 minute timer", "start a timer for 20 minutes"
* A visual is displayed in the doorhanger when being set
* Only one timer can be set at a time.
* When timer is complete the doorhanger opens with an open mic to allow for follow up commands such as "stop" or "reset." See #962 for details on follow up utterances.
* If a user quits Firefox when a timer is active we should warn them the timer will be lost.

![Timer-doorhanger](https://user-images.githubusercontent.com/428245/76723969-79d9c100-6706-11ea-8a33-5f192594b784.jpg)

[zap-circle.svg.zip](https://github.com/mozilla/firefox-voice/files/4336084/zap-circle.svg.zip)

awallin:
@danielamormocea this one is a bit more advanced. Might be a good fit for you, if you're interested?

GangaChatrvedi:
@awallin I was already working on this if that is alright.

danielamormocea:
@awallin  yes, very; I was actually thinking about it last week

GangaChatrvedi:
@danielamormocea Okay. Please go on.

danielamormocea:
@awallin  I was thinking about making something like this for checking the weather (of the current day as the prompt is fairly small) also? If you consider this a useful feature I could make an issue and work on that too;

awallin:
@danielamormocea Yes, a native weather feature would be nice. I'll put together a mockup and we can see who is available to work on it.

danielamormocea:
@ianb  @awallin  How should I announce that the timer has finished counting? Should I use a jingle? Or a warning?

awallin:
@danielamormocea There's a mockup showing the timer has finished. The one with "It's been 5 minutes"

danielamormocea:
@awallin @ianb Should we allow the user to have more than one timer? 
And if so:
-> the user should be able to see all of the timers? (as in sliding to see the other timers; right-left arrows)
[...]

PR #1296, opened 2020-03-16
WIP: Allow followup questions

### LINKED ISSUE #962: Allow followup questions
In some cases we want to ask for followup from the user (usually "yes" or "no", but maybe "next" or some other phrase). Right now they'd have to start the intent all over.

For this ticket we'd allow an intent to specify that it would like to receive another response from the user, and that would get fed back in. If the user closes the popup then it's cancelled.

JM-Mendez:
This PR will resolve #962 which allows listening for follow up questions.

#### Next intent/phrase directly associated with previous intent
1. user says "search for x"
2. show search results and listen for follow up
3. user says "next result"
4. the calling intent handles the follow up
5. possibly listen for more follow ups
___

### Proposed Solution: (**EDITED**)
My plan of attack is creating a method on the context named `.listenForFollowup()` that sets a global file variable named `runFollowUp`. 

On the next utterance, I check if a follow up has been requested. If it is, I match the the utterance to the list of follow up phrases, rather than the default ones.

I then call `intent.runFollowUp(nextContext)`. The new context will have a `followUpLabel` property that will allow the intent to know which follow up phrase was matched.

Implementing the UI will be engineered after implementing the follow up logic.

### Tasks:
- [x] set method on the context named `.listenForFollowUp()`
- [x] parse specified follow ups per intent
- [x] call `.runFollowUp()` if the a follow up is requested
- [x] implement UI components


JM-Mendez:
@ianb can you please give this a quick once over? I'm concerned my proposal may have increased the ticket's scope, and also want to ensure that the proposed solution is extensible for future requirements.

JM-Mendez:
Based on open tickets #1311 and #76, I think that allowing non-related intents to handle follow ups should be implemented after related intents. Thoughts?

ianb:
@JM-Mendez Note you don't have any commits in this PR beyond the base search-related commits, so I don't think there's anything for me to look at.

For the first version of this, yes, we should only allow intent-specific followups. So the intent itself would ask for followup (via some method on `IntentContext`) and would process the followup utterance. We can think through more unrelated followups later.

extension/background/intentParser.js
```diff
@@ -20,34 +20,65 @@ export function registerMatcher(intentName, matcher) {
...
+            entities: entityTypes,
+            intentName: name,
+          });
+
+          compiled.followUpLabel = phrase.label;
+          compiled.followUpUtterance = true;
+          return compiled;
+        });
+      });
+    }
```
JM-Mendez: This is my main concern. Since this section is already wrapped in a loop in the upper scope, compiling the follow up phrases has a complexity of n3 (yikes!)....

JM-Mendez:
Hi @ianb, yeah sorry about the miscommunication. I was asking just for your input on the proposal I made. It's alright though because I ended up going in a slightly different direction, since my initial idea ended up having to refactor way too much.

I've pushed up some changes that I think are a bit simpler. Although I don't have enough context on the codebase yet to know for sure. I haven't implemented the UI component yet, so the only way to test it is manually. 

If you'd like to test it, I've pushed up a temporary commit that runs follow up search commands. At the moment it works as before the changes, except that it if you say "next result" or "previous result", it will log to the console that a follow up command was run. Also, the `console.log` statements are why the tests are failing.

I'll move onto the rest of the implementation while I await your input.

extension/background/intentParser.js
JM-Mendez: I've addressed this in commit fba590fd7bfc20cb9a892e183e421da1d6ca5e52 and reduced the complexity to n^2, which is what it was before this PR

JM-Mendez:
Hi @ianb, this is my final take on adding follow ups. Currently the only intent I have working is the `search.searchPage` ("search for x", "next", "previous"), and that's just as an example. Since I had to touch many files, do you think I should add the actual `FollowUp` component in another issue?

The `Success` component is currently not wrapped in a follow up, because I wasn't sure if @awallin wanted a new success view when listening for a follow up. 

But I made this into a component similar to `IntentFeedback`, which can be added at will.

Here it is in action. 
movie

___
And @awallin, here is a static shot of the component in case you want to give me notes
Screen Shot 2020-03-20 at 19 43 01

extension/background/intentRunner.js
```diff
@@ -266,18 +279,26 @@ export async function runUtterance(utterance, noPopup) {
...
+    runFollowUp,
+    lastIntentName: lastIntent && lastIntent.name,
+  });
   desc.noPopup = !!noPopup;
   return runIntent(desc);
 }
 
 export async function runIntent(desc) {
+  runFollowUp = runFollowUp && desc.followUpLabel;
+  if (lastIntent && runFollowUp) desc.name = lastIntent.name;
```
ianb: nit: There should be braces here

extension/intents/search/search.toml
```diff
@@ -80,6 +80,17 @@ match = """
...
+[[search.searchPage.followUp]]
+label = 'nextResult'
+match = """
+  (search |) next (to view |) (search |) (result{s} | item{s} | page | article |)
+""" 
+
+[[search.searchPage.followUp]]
+label = "previousResult"
+intent = 'search.previous'
+
```
ianb: Is the idea in the second example here that it just takes the phrases from `search.previous` instead of listing them?

extension/popup/popupController.jsx
```diff
@@ -396,10 +399,27 @@ export const PopupController = function() {
...
+        });
+
+        waitForFollowUp = runFollowUp;
+        if (runFollowUp) {
+          recorder.startRecording();
+          runFollowUpTimerId = setTimeout(async () => {
+            await browser.runtime.sendMessage({
+              type: "stopListeningForFollowUp",
+            });
+          }, 5000);
```
ianb: Is the idea here that after 5 seconds it stops listening for a followup?

The recorder also has a silence timeout that might be the right thing to use here.

extension/popup/popupView.jsx
```diff
@@ -639,27 +643,38 @@ const SearchResultsContent = ({
...
-              {new URL(next.url).hostname} | {next.title}
-            </a>
-          </div>
-        ) : null}
+        <FollowUpContainer>
+          {next ? (
+            <div id="next-result">
+              <p>
+                <strong>
+                  Click mic and say <i>'next'</i> to view
```
ianb: "Click mic" could be removed here I believe

awallin:
In general, I think it makes sense what your saying @ian. Any intent can be execute while the mic is left open but priority is given to certain intents (due to their context) we expect (yes, no, cancel, next, etc).

As a first pass I think there are some intents that should just always listen for a follow up. The mic will stay open for x seconds allowing that follow up to be input. This feels like something will adjust as we play with it and get user feedback

JM-Mendez:
>I can't decide what exactly I think about how phrases should be parsed. Right now using your code if I do "search lions" then "new tab" it will search for "new tab", which doesn't seem right at all. I'd either want it to reject "new tab" or else run the matching intent.

@ianb, you are correct that it's not right. The intention was for it to run the matching intent. After your notes about wanting to let any intent run, but have certain matches run as as followup (yes/no/etc), I'll have to throw away most of my changes. So I'll resolve it then.

I'm also wondering if you want the mic to stay open if the popup isn't closed, and if that should be the default behavior. Currently all intents default to closing upon completion, and that's why I had the timeout to close it after x seconds. 

We can make it the default to stay open, or the dev can decide to explicitly listen for follow ups. 

Maybe @awallin has some insight into the preferred UX?

docs/writing-an-intent.md
```diff
@@ -39,6 +39,16 @@ match = """
   a{n} (test | example) intent
 """
 
+[[someIntent.command.followUp]]
+label = "A label for this follow up phrase, to differentiate between follow ups"
```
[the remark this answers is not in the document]
JM-Mendez: Yes, it was supposed to be like an ID. This change will become irrelevant in the next iteration, since we're moving in a slightly different path.

extension/background/intentParser.js
```diff
@@ -20,34 +20,73 @@ export function registerMatcher(intentName, matcher) {
...
+      phrases.push({ compiled });
+    }
+
+    if (followUp) {
+      followUpPhrases[name] = [];
+      // Some intents may need be able to return to the base intent.
+      // For example, `search.searchPage` can be followed up by `search.next`.
+      // A user may then wish to execute another search before the popup closes,
+      // but it will be treated as a follow up, when it should reset to the base intent
+      followUpPhrases[name].push({ compiled: INTENTS[name].compiledMatcher });
```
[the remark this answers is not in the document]
JM-Mendez: Originally I needed a way to allow the same intent to be called as a follow up. This will no longer be relevant in the next iteration.

extension/background/intentParser.js
```diff
@@ -20,34 +20,73 @@ export function registerMatcher(intentName, matcher) {
...
+
+      for (const nextPhrase of followUp) {
```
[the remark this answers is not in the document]
JM-Mendez: Got it. Although in the next iteration I'll be storing the compiled follow up matchers directly in the intent as you've asked.

extension/background/intentParser.js
```diff
@@ -20,34 +20,73 @@ export function registerMatcher(intentName, matcher) {
...
+        const desc = {
+          followUpLabel: nextPhrase.label,
+          followUpUtterance: true,
+        };
+
+        if (nextPhrase.intent) {
```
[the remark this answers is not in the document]
JM-Mendez: The initial implementation would allow an intent to set a different intent as follow up. Now that any intent can be called while the mic is open, this will no longer be required.

awallin:
For privacy reasons we should never allow the mic to be open unless the popup is also open and it's clear the mic is listening. (Exception is if the users has enabled wakeword)

extension/background/intentParser.js
```diff
@@ -20,34 +20,73 @@ export function registerMatcher(intentName, matcher) {
...
+          }
+        }
+
+        followUpPhrases[name].push(desc);
+      }
     }
   }
-  phraseSet = new PhraseSet(phrases);
+
+  phraseSet = new PhraseSet(phrases, followUpPhrases);
```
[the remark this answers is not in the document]
JM-Mendez: This makes sense. Having been studied the codebase more, I realized that if an intent only has a phrase matcher, then I don't have to touch `PhraseSet`. I'll revert in the next iteration.

extension/intents/search/search.toml
JM-Mendez: Yes, that was the idea. But with allowing any matching intent be run in it's own handler, and only direct follow up phrases handled as a follow up, I think this is no longer necessary.

extension/popup/popupController.jsx
JM-Mendez: This would complete the intent and close the pop up window, since it was being held open. I could move this to the context as well, but thought the user might want to set the timer in their settings.

JM-Mendez:
> For privacy reasons we should never allow the mic to be open unless the popup is also open and it's clear the mic is listening. (Exception is if the users has enabled wakeword)

@awallin got it. Should we default to always keeping the pop up open for "x seconds", with the new "mic is on" component? Currently successful intents force close the pop up, unless explicitly asked to stay open in the intent.

awallin:
@JM-Mendez Yes, default to keep the pop up open for x seconds (maybe 5 seconds?) when the "mic is on" component has been invoked. Maybe different intents can extend the time but by default it's set to the minimum we think is proper amount of time for a user to identify, read, and give an utterance if they'd like.

ianb:
Cool, sounds like we have a good plan here. Even just playing with this version for 10 minutes was helpful in thinking about how it should work, and I expect we'll learn more if we land it and let it take any intent input, and see whether that feels weird or cool. We want to do everything we can to encourage people to use the tool more, so making it listen more and listen for anything is a good place to start.

ianb:
For a followup (@awallin design needed): we should definitely animate the indicator that the mic is on.

ianb:
@JM-Mendez is this ready for a final review? (It's still in draft)

JM-Mendez:
@ianb here is the next iteration for allowing follow ups. This time I created two versions for you to play with:

1. per intent (default setting)

   - will listen for follow ups only when requested
   - will throw an error if a follow up is not understood, and allows retries.

2. global (set in extension settings)

   - will listen for follow ups on all intents
   - an intent with a follow up has priority to run
   - an intent can insist on a follow up response (see flows below)

### Flows:

#### Per intent (default setting)

Bookmarks (only `bookmarks.remove` has a follow up)

-  say "bookmark this page"
-  say "remove bookmark"
-  say "yes" or "cancel" (valid phrase)
-  say any other phrase or intent (invalid but allows retries)

Searching (only `search.searchPage` has a follow up)

-  say "search for xyz"
-  say "next" or "previous" (valid phrase)
-  say any other phrase or intent (invalid but allows retries)

#### Global

Bookmarks

- Same as above, but will not redirect on different intents (it insists on follow up responses)

Searching 

- Same as above, but **_will_** redirect on different intents



___
What this doesn't cover yet:
- Default followup text (can be improved?)
- text input component
- search result clickable links (I'll have to move follow up hints from context to components)
- Should I handle error and/or logging?

JM-Mendez:
@ianb I forgot to mention that since there are many commits, if you choose to merge these changes I can squash/rebase locally so there could be a more detailed commit message in the master branch.

extension/background/intentRunner.js
```diff
@@ -89,6 +112,68 @@ export class IntentContext {
...
+      };
+    }
+    return browser.runtime.sendMessage({
+      type: "handleFollowup",
+      method: "enable",
+      message: {
+        heading: this.followupHint.heading,
+        subheading: this.followupHint.subheading,
+      },
+    });
```
ianb, line 133: Something about this part doesn't seem to be working for me now. The popup doesn't reopen the mic and doesn't show a message, even though it worked previously. 

Actually I'm getting all sorts of peculiar behavior. If I do "search for lions" then it shows the followup, but doesn't actually work. Or, maybe the initial timeout is too low, so when I saw "next" the recorder has already stopped (is there an indicator?)

Then the search gets a little wonky as well, it seems to lose track of the search page in a bunch of cases, and I get missing message handler messages.

extension/intents/bookmarks/bookmarks.js
```diff
@@ -76,20 +76,38 @@ intentRunner.registerIntent({
...
+          context.endFollowup();
+          const e = new Error("This page wasn't bookmarked");
+          e.displayMessage = "This page wasn't bookmarked";
+          throw e;
+        }
 
-    await browser.bookmarks.remove(selected[0].id);
+        await browser.bookmarks.remove(selected[0].id);
+        await context.displayText("Bookmark has been removed");
+        await context.endFollowup();
```
ianb, line 110: But, for whatever reason the bookmark stuff worked well for me, including listening for a fairly long time.

JM-Mendez:
@ianb since the bookmarks intent works, but the weird issues popping up happen in the search intent, it might have to do with a PR that hadn't been merged when I pushed up these changes (#1314). I've rebased off the latest master, so hopefully that behavior does reoccur.

extension/background/intentRunner.js
```diff
@@ -273,8 +362,32 @@ export async function runUtterance(utterance, noPopup) {
...
+      lastIntentForFollowup.acceptFollowupIntent.includes(desc.name)
+    ) {
+      desc.isFollowupIntent = true;
+    } else if (
+      !globalListenForFollowup ||
+      (globalListenForFollowup && lastIntentForFollowup.insistOnFollowup)
+    ) {
+      lastIntentForFollowup.failed(
+        `The phrase\n\n"${desc.utterance}"\n\nis invalid. Please try again`
+      );
```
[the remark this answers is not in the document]
JM-Mendez: see https://github.com/mozilla/firefox-voice/pull/1296#issuecomment-607524818.

extension/background/intentRunner.js
```diff
@@ -17,10 +18,15 @@ const METADATA_ATTRIBUTES = new Set([
...
   "match",
+  "followup",
 ]);
 
 export const intents = {};
 let lastIntent;
+// This is a copy of lastIntent that gets nullified when the pop up
+// window closes.
+let lastIntentForFollowup;
+let globalListenForFollowup = settings.getSettings().listenForFollowup;
```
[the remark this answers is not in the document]
JM-Mendez: I put this in the preferences view. Is that not the right place? I put it there because 1) I wasn't sure if you'd want continuous followups and 2) I thought the user may get to decide

Here's a screenshot for reference
Screen Shot 2020-04-01 at 6 41 10 PM

extension/background/intentRunner.js
```diff
@@ -30,10 +36,26 @@ db.createTable(utteranceTable, primaryKey, voiceVersion)
...
 
+settings.watch("listenForFollowup", ns => {
+  globalListenForFollowup = ns;
+});
+
 export class IntentContext {
   constructor(desc) {
     this.closePopupOnFinish = true;
     this.timestamp = Date.now();
+    this.followupPhraseSet;
```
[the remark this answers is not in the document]
JM-Mendez: I'll fix this. It's a remnant from an earlier thing I did.

extension/background/intentRunner.js
JM-Mendez: see https://github.com/mozilla/firefox-voice/pull/1296#issuecomment-607524818, it might fix this. Although not sure what you mean about the missing message handler messages.

JM-Mendez: As for an indicator, we don't have one yet. Currently the timer is set to close the popup after 5 seconds. One issue I ran into with the search intent was how to start the timer *after* the page loaded. I could only get the search results page to wait until load, but saying "next/previous" would not.

ianb:
When I more carefully applied this it seemed to work properly – think there's some problem with jsx files not always rebuilding properly, even when closing the browser and running npm start. I had the same problem reviewing @danielamormocea's timer PR.

I think there's probably some regressions caused by this all, but I definitely like how it feels and it'll be easier to find and fix those in master than in a PR. 

Very nice work, and I think it will have a real effect on the feel of the product!

PR #1300, opened 2020-03-16
RE: Close active tab alias

### LINKED ISSUE #1298: Close active tab alias
As reported by a user "close active tab" and "close the active tab" should work as an alias for "close tab"

xlisachan:
Fixes #1298

I noticed the word "active" was not supported/included in the "tabs.close" intent, please review and let me know if it is necessary. If it is not, I will close this pull request. Thanks!


ianb:
Cool, aliases like these are often helpful. The examples should have `test=true`, but I'm just going to add that in real quick...

PR #1305, opened 2020-03-16
Create start slideshow intent

### LINKED ISSUE #1115: Intent: start slideshow
This would try to find images and/or videos on the page that would be suitable for a slideshow.

This extension might offer some inspiration: [Slideshow](https://addons.mozilla.org/en-US/firefox/addon/media-slideshow/)

dave-ok:

Fixes #1115
This is a first draft for the intent to start a slideshow of images and videos in a tab. Slideshow design is minimal, but may need more work. Tests are currently failing locally after a recent fork sync, particularly selenium tests I think. @ianb please kindly review.

### DIFF
--- a/extension/intents/slideshow/contentScript.js
+++ b/extension/intents/slideshow/contentScript.js
@@ -1,0 +1,155 @@
+/* globals communicate */
+
+this.slideshowScript = (function() {
+    let slideElements = [];
+    let currentSlideIndex = 0;
+
+    const videoSources = [
+        "youtube.com",
+        "vimeo.com"
+    ];
+
+    slideElements = detectAllMedia();
+    buildSlideStructure();
+
+    const lbContainer = document.body.querySelector(".fv-lightbox-container");
+    const slideContainer = document.body.querySelector(".fv-slide-image");
+
+    function buildSlideStructure() {
+        const lightboxElement = document.createElement("div");
+        lightboxElement.className = "fv-lightbox-container";
+        lightboxElement.onclick = hideSlideShow;
+
+        const slideshowContainer = document.createElement("div");
+        slideshowContainer.className = "fv-slideshow-container";
+
+        const slideImage = document.createElement("div");
+        slideImage.className = "fv-slide-image";
+
+        const tagPrev = document.createElement("a");
+        tagPrev.className = "fv-prev";
+        tagPrev.textContent = String.fromCharCode(10094);
+        tagPrev.onclick = previousSlide;
+
+        const tagNext = document.createElement("a");
+        tagNext.className = "fv-next";
+        tagNext.textContent = String.fromCharCode(10095);
+        tagNext.onclick = nextSlide;
+
+        slideshowContainer.appendChild(slideImage);
+        slideshowContainer.appendChild(tagPrev);
+        slideshowContainer.appendChild(tagNext);
+
+        lightboxElement.appendChild(slideshowContainer);
+
+        document.body.appendChild(lightboxElement);
+    }
+
+    function hideSlideShow(event) {
+        if (event.target === lbContainer) {
+        lbContainer.style.display = "none";
+        swapSlide();
+        }
+    }
+
+    function revealSlideshow() {
+        swapSlide(slideElements[0]);
+        lbContainer.style.display = "block";
+    }
+
+    function isVideoSourceUrl(url) {
+        let testExp = "";
+        const result = videoSources.some(function(videoSource) {
+            testExp = new RegExp(`^(https?:\/\/)?(www\.)?${videoSource}`);
+            return testExp.test(url);
+        });
+
+        return result;
+    }
+
+    function detectAllMedia() {
+        const selector = "img, video, iframe";
+        // detect all supported elements
+        const mediaElements = document.body.querySelectorAll(selector);
+
+        // iterate thru list and add to array
+
+        mediaElements.forEach(element => {
+            // strip their current classes
+
+            element.className = "";
+            switch (element.tagName) {
+                case "IMG": {
+                    const img = document.createElement("img");
+                    img.setAttribute("src", element.getAttribute("src"));
+                    slideElements.push(img);
+
+                    break;
+                }
+                case "VIDEO": {
+                    const vid = document.createElement("video");
+                    vid.setAttribute("src", element.getAttribute("src"));
+                    vid.setAttribute("controls", true);
+                    slideElements.push(vid);
+
+                    break;
+                }
+                case "IFRAME": {
+                    if (isVideoSourceUrl(element.getAttribute("src"))) {
+                        const iframe = document.createElement("iframe");
+                        iframe.setAttribute("src", element.getAttribute("src"));
+                        iframe.setAttribute("width", 640);
+                        iframe.setAttribute("height", 400);
+                        slideElements.push(iframe);
+                    }
+
+                    break;
+                }
+            }
+        });
+
+        return slideElements;
+    }
+
+    function previousSlide() {
+        moveSlide(-1);
+    }
+
+    function nextSlide() {
+        moveSlide(1);
+    }
+
+    function moveSlide(steps) {
+        let newIndex = currentSlideIndex + steps;
+        if (newIndex > slideElements.length - 1) {
+            newIndex = 0;
+        }
+        if (newIndex < 0) {
+            newIndex = slideElements.length - 1;
+        }
+
+        showSlide(newIndex);
+    }
+
+    function showSlide(slideIndex) {
+        currentSlideIndex = slideIndex;
+        swapSlide(slideElements[slideIndex]);
+    }
+
+    function swapSlide(newSlide) {
+        if (slideContainer.hasChildNodes()) {
+        slideContainer.firstChild.remove();
+        }
+        if (newSlide) {
+        slideContainer.appendChild(newSlide);
+        }
+    }
+
+    communicate.register("openSlide", async (message) => {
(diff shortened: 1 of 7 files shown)


dave-ok:
@ianb I have made a few modifications to slideshow structure to better emulate the sample slideshow structure

extension/intents/slideshow/contentScript.js
```diff
@@ -0,0 +1,186 @@
...
+        slideContainer.firstChild.setAttribute(
+          "height",
+          document.documentElement.clientHeight
+        );
+      }
+    }
+  };
+
+  function buildSlideStructure() {
+    const lightboxElement = document.createElement("div");
```
ianb: There's a danger this could conflict with the underlying page, but that's something that could be fixed in a followup.

But to give a hint, how this would generally be fixed would be something like:

```js
const iframe = document.createElement("iframe");
iframe.src = browser.runtime.getURL("/intents/slideshow/slideshow.html");
iframe.style.position = "absolute"; // and some other positioning styles
document.body.appendChild(iframe);
```

And you can manipulate the contents of the iframe much like you would the original document, just using `iframe.contentDocument` instead of `document`.

`slideshow.html` could also have its own stylesheet referenced from the file itself.

extension/intents/slideshow/contentScript.js
```diff
@@ -0,0 +1,186 @@
...
+
+    const tagClose = document.createElement("a");
+    tagClose.className = "fv-close";
+    tagClose.textContent = String.fromCharCode(0x274c);
+    tagClose.onclick = hideSlideShow;
+
+    const tagPrev = document.createElement("a");
+    tagPrev.className = "fv-prev";
+    tagPrev.textContent = String.fromCharCode(10094);
+    tagPrev.onclick = previousSlide;
```
ianb: It doesn't _really_ matter, but it's considered better form to use `tagPrev.addEventListener("click", previousSlide);`

extension/intents/slideshow/contentScript.js
dave-ok: Thanks @ianb for your recommendations. By conflicts, I'm guessing you mean stylesheets from various sites could end up disrupting the slideshow's structure? I didn't know I could reference and use `html` files in this way, would've made things easier to visualize, easier to implement. I'll get to work on these right away

dave-ok: @ianb I have modified the slideshow to use an iframe as you advised. However, this has brought on a challenge. At the point where I try to append the new Iframe object with given src to the document object, I get this security error "Security Error: Content at http://xxxx.xxx/xxxx may not load or link to moz-extension://a669dc66-0200-48e7-83da-e7a815fac9f5/intents/slideshow/slideshow.html". From observation this seems to happen when the Iframe's src is being parsed. I felt this was a permissions issue. I tried to add the slideshow.html to "web_accessible_resources" in manifest.json. Still no changes. Please, I need your help in moving forward from this point

dave-ok: @ianb Finally got the slideshow into an iframe (TIL Iframe's are a pain). Got past cross domain issue/permissions by skipping src altogether. And found out much hair-pulling that contentDocument is only accessible after "onload" event. Thanks for these suggestions I have learned a lot

extension/intents/slideshow/contentScript.js
dave-ok: @ianb I have modified all event handlers to use this form

dave-ok:
@ianb I have used an Iframe to render the slideshow. I am moving forward with the thumbnail gallery

dave-ok:
@ianb Please, when you have the time can you review this pull request? I have completed the thumbnail gallery and the slideshow section. No interference with a couple of sites that I have tested. Only challenge/caveat is that some sites might not render some videos because they have obscured the link to their videos. A case in point is reddit videos. After some searching found out this is particular to them and is most likely by design to drive traffic to their site. I don't know if this is sufficient or if you have any further suggestions. I will appreciate your thoughts on this. Thanks

ianb:
I tried it on this and it got weird: https://imgur.com/gallery/cJGrYKq

image

But then on pinterest it seemed similar: https://www.pinterest.com/pin/514465957436689055/

image

But on a search of squirrels it seemed better: https://www.google.com/search?hl=en&tbm=isch&source=hp&biw=1280&bih=953&ei=JHt6XrmPF4i4tAbg-bfoDQ&q=squirrels&oq=squirrels&gs_l=img.3..0l10.4242.5304..5463...0.0..1.181.999.1j7......0....1..gws-wiz-img.......0i131.wJ99mbvANro&ved=0ahUKEwi5sf_LhrToAhUIHM0KHeD8Dd0Q4dUDCAU&uact=5

It only gets the thumbnails, which is kind of inevitable (at least for now). A wikipedia page worked better: https://en.wikipedia.org/wiki/Hours_of_Mary_of_Burgundy

This page would work well except for some terrible lazy loading thing that seems to foil every off-page image: https://healthyhandyman.com/woodworking-projects-for-kids-to-build/

extension/intents/slideshow/contentScript.js
```diff
@@ -0,0 +1,347 @@
...
+    // create iframe element
+    iframeContainer = document.createElement("iframe");
+    iframeContainer.className = "fv-slideshow-frame";
+    iframeContainer.height = document.documentElement.clientHeight;
+    iframeContainer.width = document.documentElement.clientWidth;
+    iframeContainer.style.display = "none";
+    iframeContainer.style.position = "fixed";
+    iframeContainer.style.top = 0;
+    iframeContainer.style.left = 0;
+    iframeContainer.style.zIndex = 100;
```
ianb: This zIndex should be much higher, as it often ends up underneath other page components. In Screenshots we use 99999999999

Also the iframe should have a `src`. I'd create a basically blank document and use `iframe.src = browser.runtime.getURL("/intents/slideshow/blank.html");` By making our own document we keep the containing page from accessing it.

extension/intents/slideshow/contentScript.js
```diff
@@ -0,0 +1,347 @@
...
+
+    document.body.appendChild(iframeContainer);
+
+    iframeContainer.addEventListener("load", function() {
+      // contentWindow is first accessible here
```
ianb: And, a very obscure security check that pertains only to iframes created by webextensions – here you want to verify that the iframe still has the same src, like `if (iframeContainer.documentElement.documentURI !== browser.runtime.getURL("...")) {throw new Error(...)}`

extension/intents/slideshow/contentScript.js
```diff
@@ -0,0 +1,347 @@
...
+    const mediaElements = document.body.querySelectorAll(selector);
+
+    // iterate thru list and add to array
+
+    mediaElements.forEach(element => {
+      // strip their current classes
+
+      element.className = "";
+      switch (element.tagName) {
+        case "IMG": {
```
ianb: I think there should be a minimum size for images (probably looking at clientHeight/Width). E.g., at least 200px in both directions. As it is now it will pick up some controls and icons.

extension/intents/slideshow/contentScript.js
```diff
@@ -0,0 +1,347 @@
...
+        // enable iframe to resize first
+        newSlide.setAttribute("width", document.documentElement.clientWidth);
+        newSlide.setAttribute("height", document.documentElement.clientHeight);
+      }
+      slideContainer.appendChild(newSlide);
+    }
+  }
+
+  communicate.register("openSlide", async message => {
+    revealSlideshow();
```
ianb: Right now it starts on the overview, but I think showing the first slide would be clearer.

Also if you could catch ArrowLeft and ArrowRight and go forward/back on slides it would be helpful. I probably was confused because there was side decoration over the buttons the first time I tried it, but I naturally expected keys to work.

extension/intents/slideshow/slideshow.toml
```diff
@@ -0,0 +1,12 @@
+[slideshow.open]
+description = "Opens the slideshow"
+match = '''
+    (open | display | launch | start) (the |) (slideshow | slide shoe)
```
ianb: The first word could be optional, i.e., "slideshow" alone could start a slideshow.

You could put "slide shoe" into `english.toml`, but it's fine here too.

extension/intents/slideshow/contentScript.js
dave-ok: I have adjusted the z-index and added the blank html file

extension/intents/slideshow/contentScript.js
dave-ok: Added the security check

extension/intents/slideshow/contentScript.js
dave-ok: Set a minimum size for images of 200x200px

extension/intents/slideshow/contentScript.js
dave-ok: Made first slide the default view. Added Support for Arrow keys and escape key

extension/intents/slideshow/slideshow.toml
dave-ok: Made "slideshow" only enough to start slideshow. Moved the alternate spelling to english.toml

dave-ok:
@ianb Please when you have the time could you review this? Turns out image lazy loading on some sites is caused by jquery plugins. Found a way around it. I also implemented functionality to allow manually reloading/updating the slideshow gallery in case the workaround fails or for sites with infinite scrolling and lazy loading

ianb:
Thanks!

PR #1314, opened 2020-03-17
Always opens a new tab on search intent

### LINKED ISSUE #1246: Google tab performs search for "Search" commands meant to open new result tabs
### Prerequisites:
Mic permissions are enabled.
"A privacy reminder from Google" is reviewed and agreed and the yellow language pop-up is closed.
A Google tab is opened.

### STR:
1. Use the shortcut or click on the mic icon from the browser toolbar in a tab, other than Google.
2. Say or write a "Search" command meant to open a new tab, e.g. "What movies are playing at the cinema", "How to make fish tacos".

### Expected result:
A new tab opens with the result, without interfering with other opened tabs.

### Actual result:
The Google tab performs a search for the command made.

### Notes:
Reproduced on Mac 10.14 and Win10x64 with Firefox Nightly 76.0a1 (2020-03-11).
After performing the search, the Google tab sometimes closes and it's not shown in the hidden tabs list.
Making the command, from step 2, in the Google tab has the same result.

shreyaa-s:
# fixes #1246 

refactors openSearchTab() to always open new tab on search intent
removes closeSearchTab()

I have refactored `openSearchTab()` to always open a new tab on search intent.

### DIFF
--- a/extension/intents/search/search.js
+++ b/extension/intents/search/search.js
@@ -6,15 +6,12 @@ import * as content from "../../background/content.js";
 import * as telemetry from "../../background/telemetry.js";
 import * as browserUtil from "../../browserUtil.js";
 
-// Close the search tab after this amount of time:
-const CLOSE_TIME = 1000 * 60 * 60; // 1 hour
 // Hide the search tab after this amount of unfocused time:
 const HIDE_TIME = 1000 * 60; // 1 minute
 // Check this often for a new image in a search tab:
 const CARD_POLL_INTERVAL = 200; // milliseconds
 // If we don't believe the card is animated, still get new images for this amount of time:
 const CARD_POLL_LIMIT = 1000; // milliseconds
-let closeTabTimeout = null;
 let lastImage = null;
 let _searchTabId;
 const START_URL = "https://www.google.com/?voice";
@@ -27,23 +24,8 @@ const tabSearchResults = new browserUtil.TabDataMap(5 * 60 * 1000); // only dele
 let googleIsDefaultProvider;
 
 async function openSearchTab() {
-  if (closeTabTimeout) {
-    clearTimeout(closeTabTimeout);
-    closeTabTimeout = null;
-  }
-  closeTabTimeout = setTimeout(() => {
-    closeSearchTab();
-  }, CLOSE_TIME);
-  if (_searchTabId) {
-    try {
-      await browser.tabs.get(_searchTabId);
-      return _searchTabId;
-    } catch (e) {
-      // Presumably the tab doesn't exist
-      log.info("Error getting tab:", String(e));
-      _searchTabId = null;
-    }
-  }
+  _searchTabId = null;
+
   for (const tab of await browser.tabs.query({
     url: ["https://www.google.com/*"],
   })) {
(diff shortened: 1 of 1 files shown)


shreyaa-s:
@JM-Mendez is this what is required?

JM-Mendez:
@shreyaa-s My apologies. In an effort to not give you the answer directly, I was unfortunately too vague with my description.

The bug is localized to the `search.searchPage` intent within these lines:
https://github.com/mozilla/firefox-voice/blob/c16b3d06041d6973221e050c8595db7565b7d62c/extension/intents/search/search.js#L366-L374

I mistakenly used `openSearchTab()`. If you look at the implementation of that function, it will create a search tab if one doesn't exist, and reuse a tab if it does.

Instead, you will need to create a new tab there. If you look at the file [`intentRunner.js`](https://github.com/mozilla/firefox-voice/blob/c16b3d0604/extension/background/intentRunner.js), in the `IntentContext` class you will find a method to create a tab. The context is the parameter you'll receive in the intent. You will need to pass a configuration object to the method, even if its an empty object.

From the return value you can retrieve the `tabId`. You can then pass that id to `browser.search.search`. You will also have to pass a starting url to `browser.seach.search`, since you will be explicitly giving it a tab to use.

Lastly you can remove `await focusTab()` since it won't be required at that point. 

If you get stuck, reach out.

shreyaa-s:
@JM-Mendez since it had a query whose working wasn't clear to me, i decided not to interfere with `openSearchTab()` and returned a null value in `_searchTabId` so that it creates a new tab everytime a search intent is called.

JM-Mendez:
@shreyaa-s I offer my advice with the caveat that I'm an not a mentor. Just wanted to make that clear in case you preferred to hear directly from one. I definitely wouldn't feel offended if you pinged them directly ☺️ 
___

If you search for `_searchTab` in that file, you'll notice it is being used in multiple places. Most importantly in another intent named `search.show`

https://github.com/mozilla/firefox-voice/blob/c16b3d06041d6973221e050c8595db7565b7d62c/extension/intents/search/search.js#L345-L348

I can't say for sure, but it looks like resetting that variable would most likely break other parts of the codebase.

What would be preferable I think, is to set the newly created `tabId` to `_searchTabId`. This way the other areas can work as expected.

You did bring my attention to another bug that would arise from these changes tho. Here I grabbed the results from the last saved `_searchTabId`. 

But if the user can now have multiple search tabs open, you'll have to make sure to grab the correct tab id, which is most likely the currently active tab.

https://github.com/mozilla/firefox-voice/blob/c16b3d06041d6973221e050c8595db7565b7d62c/extension/intents/search/search.js#L186-L192

At that point it would be beneficial to throw an error (it will display to the user), if they use this intent on a non-search tab.

You can model your error similar to the other's in that area
https://github.com/mozilla/firefox-voice/blob/c16b3d06041d6973221e050c8595db7565b7d62c/extension/intents/search/search.js#L180-L184
___
Again, feel free to ping a mentor directly if you'd rather hear from them.

shreyaa-s:
@JM-Mendez I'm making the final changes just had a few doubts regarding the intentions for the implementation of previous and next result intents. Is this feature solely for last opened search tabs? Also the race between `closeSearchTab()` and `_tabHideListener()` can you shed some light upon it?

JM-Mendez:
@shreyaa-s I replied to your question in the chat room

shreyaa-s:
@ianb review this when you get a moment. Thanks!

shreyaa-s:
hey @ianb can you have a look at this when you have a moment?

PR #1295, opened 2020-03-16
RE: Converted test/test.js to a Jest test

xlisachan:
Fixes #1255 

I have renamed the test `integration.test.js`, but used focusing tests (describe(),  it()) instead of wrapping it with test(), expect() since it is locating the button on the toolbar. 

I also had to add a timeout to solve an error that said `async callback was not invoked within the 5000ms timeout`. Please review and let me know your thoughts. Thank you!

Screen Shot 2020-03-16 at 11 17 09 AM

PR #1317, opened 2020-03-17
RE: Converted test/test.js to a Jest test

### LINKED ISSUE #1255: Convert test/test.js to be a Jest test
The test in test/test.js (from #1254) is a script, I think it should become a Jest test. This might require renaming it (e.g., to `integration.test.js`) and then wrapping it in `test()` and `expect()`.

xlisachan:
Fixes #1255 
As advised, I created a fresh branch after breaking the last one #1295 😬sorry @ianb !


ianb:
Great, thanks!

Issue #992, opened 2020-02-13
Enable jest in CI


ianb:
I had to disable jest testing (`npm run jest`) in CI (by removing it from `npm test`), because CI couldn't handle the module rewrites. These work for me locally (Node v13.8.0), but CI is acting differently. I _don't_ want any module rewriting when actually using the extension (where the browser has native module support), and especially I don't want it for the old-style modules (including jsx files), it's only necessary for Node-based testing.

IkkyOdufade:
@ianb,Hi,  I'm an intern from Outreachy, I commented on issue #1134 , hoping to work on it. Do I have to fix this issue first?

ianb:
@IkkyOdufade No, this issue is separate, people can still use and test with jest locally so it's good to have instructions on it.

IkkyOdufade:
Hi @ianb. Thank you for the new issue. I’ll start working on it right away.

IkkyOdufade:
Hi @ianb I’m sorry I’ve not created a pr yet. Im working on my laptop. It would be normal again by tomorrow or next. Thank you for understanding.

IkkyOdufade:
Hi, @ianb. I've been trying to work on this and I added `npm run jest` to config.yml. This was the result, the tests passed. I know this is not solving the issue :(, so can I get more help?
![image](https://user-images.githubusercontent.com/47696080/78082086-13150280-73aa-11ea-9c4c-3d3ca9a4699e.png)

PR #1318, opened 2020-03-17
Fixes #1256 - config: integrate selenium in ci

### LINKED ISSUE #1256: Enable Selenium test in CI
The test added in #1254 would ideally be run in CircleCI. [This example](https://github.com/mozilla-services/screenshots/blob/master/.circleci/config.yml#L85-L100) might be helpful.

Note there's no special way to test the CI except to upload a PR, so it's likely to fix this will require uploading PRs over and over until it passes.

Simpcyclassy:

- [ ] Reference the issue you are fixing in at least one of your commit messages, as `Fixes #X`

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -16,6 +16,16 @@ jobs:
           name: Install Node Modules
           command: npm install
       - run:
+          name: create docker network
+          command: docker network create grid
+      - run:
+          name: install selenium-hub
+          command: docker run -d --rm -p 4444:4444 --net grid --name selenium-hub selenium/hub:3.141.59 
+          
+      - run:
+          name: install selenium-node
+          command: docker run -d --rm --net grid -e HUB_HOST=selenium-hub -v /dev/shm:/dev/shm selenium/node-firefox:3.141.59 
+      - run:
           name: Build XPI
           command: npm run package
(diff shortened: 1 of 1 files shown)


Simpcyclassy:
Please could you take a look @ianb

.circleci/config.yml
```diff
@@ -15,6 +15,13 @@ jobs:
       - run:
           name: Install Node Modules
           command: npm install
+      - run:
+          name: Download Selenium
+          command: curl -O http://selenium-release.storage.googleapis.com/3.5/selenium-server-standalone-3.5.3.jar
+      - run:
+          name: Start Selenium
+          command: java -jar selenium-server-standalone-3.5.3.jar -log test-reports/selenium.log
+          background: true
```
ianb: I think these should go town in the `test:` section (around line 38).

I also realized that because Jest isn't running in CI (#992), and the current functional tests are written with Jest, that this will make it hard to include Selenium. The issue in #992 is... somewhat mysterious to me, something about Jest and things running somewhat differently in CI (different Node version?)

Simpcyclassy: Okay fixing it right away. So I figured what was making my branches orphaned. I won't need to be closing more branhes

PR #1332, opened 2020-03-18
Fixes #1194 - "Google for X"

### LINKED ISSUE #1194: "google for X" should always open Google
Right now the `search.searchPage` intent uses the default search provider. But one of the phrases it catches is `google for X`, which would presumably always means search Google, not the default search provider. 

To test this set your default search provider to be DuckDuckGo, and then try a phrase like "google for apples".

Fixing this might be a matter of adjusting the phrases for `search.searchPage` (maybe moving those Google-specific phrases into `navigation.bangSearch`), or separating out the Google phrases and using a parameter, and choosing between `performSearch()` and `browser.search.search()` based on that parameter (in the `search.searchPage` `run()` method).

farhatcode:
Fix for: **#1194** google for X" should always open Google.

- Removes google matches and examples from `search.seacrh` and `search.searchPage` intents, to separate out google search - search.toml.

- Adds new intent named `search.searchGoogle` - search.toml

- Registers the `search.searchGoogle`intent - search.js

`Fixes #1194`

### DIFF
--- a/extension/intents/search/search.js
+++ b/extension/intents/search/search.js
@@ -410,6 +410,20 @@ intentRunner.registerIntent({
   },
 });
 
+intentRunner.registerIntent({
+  name: "search.searchGoogle",
+  async run(context) {
+    if (buildSettings.android) {
+      await performSearch(context.slots.query);
+    } else {
+      await browser.search.search({
+        query: context.slots.query,
+        engine: "Google"
+      });
+    }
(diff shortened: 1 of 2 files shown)


farhatcode:
@ianb Can you please have a look at it and give your feedback.

ianb:
Looks good, thanks!

PR #1335, opened 2020-03-19
Fixes #1327 - Copy image 

### LINKED ISSUE #1327: "Copy image" command not copying pictures in Google Docs or other related tabs
### Prerequisites:
Mic permissions are enabled.
A Google Docs tab is opened.
The tab that the user wants to copy the image from is made active.

### STR:
2. Say or write the "Copy image" command.
3. Go to the Google Docs tab.
4. Use the paste keyboard shortcut (ctrl+V for Windows or cmd+V for Mac) ("Paste" command currently not working for this tab, see #1326).

### Expected result:
Image is copied in the Google Doc successfully.

### Actual result:
Other information is copied or nothing happens.

### Notes:
Reproduced on Mac 10.14 and Win10x64 with Firefox Nightly 76.0a1 (2020-03-18).
If the user right clicks with the mouse on the image and selects "Copy Image", then it is possible to copy the picture in the desired tab, e.g. Google Docs, Google Slides, Google Sheets, https://www.calmlywriter.com/online/. ("Paste" command works for calmlywriter or Google Sheets. For Google Docs and Google Slides, the paste keyboard command should be used)

fleur101:
Fixes #1327 

- Adds handing not only base64 images
- Adds support for jpeg images in addition to png
- Fixes finding the biggest visible image 

### DIFF
--- a/extension/intents/clipboard/contentScript.js
+++ b/extension/intents/clipboard/contentScript.js
@@ -108,14 +108,16 @@ this.contentScript = (function() {
 
   types.copyImage = function() {
     const img = document.querySelectorAll("img");
-    const imgHeight = [];
-    img.forEach(function(element) {
-      if (isInViewport(element)) {
-        imgHeight.push(element.clientHeight);
+    let maxHeightIndex = 0;
+    const isValidSrc = (src) => src.startsWith("data:") || /png|jpe?g/.test(src.split(".").pop());
+    img.forEach((element, index) => {
+      if (isInViewport(element) && isValidSrc(element.src)) {
+        if (element.clientHeight > img[maxHeightIndex].clientHeight) {
+          maxHeightIndex = index;
+        }
       }
     });
-    const index = imgHeight.indexOf(Math.max(...imgHeight));
-    const url = img[index].src;
+    const url = img[maxHeightIndex].src;
     copyImage(url);
   };
 
--- a/extension/intents/clipboard/contentScript.js
+++ b/extension/intents/clipboard/contentScript.js
@@ -109,7 +109,8 @@ this.contentScript = (function() {
   types.copyImage = function() {
     const img = document.querySelectorAll("img");
     let maxHeightIndex = 0;
-    const isValidSrc = (src) => src.startsWith("data:") || /png|jpe?g/.test(src.split(".").pop());
+    const isValidSrc = src =>
+      src.startsWith("data:") || /png|jpe?g/.test(src.split(".").pop());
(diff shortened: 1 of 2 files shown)


extension/intents/clipboard/contentScript.js
```diff
@@ -108,14 +108,17 @@ this.contentScript = (function() {
 
   types.copyImage = function() {
     const img = document.querySelectorAll("img");
-    const imgHeight = [];
-    img.forEach(function(element) {
-      if (isInViewport(element)) {
-        imgHeight.push(element.clientHeight);
+    let maxHeightIndex = 0;
+    const isValidSrc = src =>
+      src.startsWith("data:") || /png|jpe?g/.test(src.split(".").pop());
```
ianb: URLs can be anything, it's best not to check the text of the URL to try to determine the type of the file. Anything in an `` should be an image of some sort.

fleur101:
@ianb I am sorry, I am a bit confused when you say you do not understand the purpose of change. This pull request tries to fix the problem that the image is not copied or the wrong image is copied - I referenced it in the commit messages and now in the pull request description as well. 

There were two bugs.

First is when image src was not base64 encoded image. When I looked into the code, I found out that the code was dealing only with base64 images, in particular, the line ` const binary = atob(url.split(",", 2)[1]);` in `dataUrlToBlob` function of `clipboard.js` file returned _undefined_ for images whose source did not start with "data:image/jpeg;base64," or a similar pattern. I then found how to do that for other than base64 images here: https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/clipboard/setImageData

The second bug was when selecting the biggest visible image. The wrong image was returned because the index in the `imgHeight ` array did not correspond to the index in `img `array since they have different sizes. I fixed the calculation.

As I understand now, I cannot rely on the url to get the image type.  I'll look into what you proposed. Thank you

fleur101:
Currently, the copying function works the following way: 
It searches for the biggest (largest height) image that is visible on the screen. Then it tries to copy the image to the clipboard either by fetching the image from the url or by using "data:" url. `browser.clipboard.setImageData(imageData, imageType` supports only jpeg/png images, so if the biggest image on the screen has any other type (gif, bmp, etc.), then the image is not copied.

This PR fixes the following parts of this process:
- Search now finds the correct image
- Copy now supports fetching images from non-"data:" type src
- Copy now supports JPEG images in addition to PNG

ianb:
Thanks, these look like good improvements

Issue #1193, opened 2020-03-09
Make `npm run build:jsx` atomic


ianb:
Right now `npm run watch-rebuild` is run whenever certain changes are noticed (.jsx, .toml, and .ejs file changes). Unfortunately it updates all of these files, and can cause a bunch of reloads.

The toml and ejs files should actually be OK, because `bin/script-utils.js` should only write files that actually change. But `babel` writes all the jsx files, and it takes a while, so web-ext might even see when it's only half-completed the task.

To fix this I think we should have babel write out (with `--out-dir`) to some temporary location `./build/`) and then move all the files into place as one command (maybe moving only changed files).

This would be a new script in `bin/` that runs babel and then moves the files.

IkkyOdufade:
Hi @ianb, can I have this?

ianb:
@IkkyOdufade You got it!

IkkyOdufade:
Hi Ian, I really dont know how to tackle this issue. Can you give me some pointers? Thank you.

ianb:
@IkkyOdufade You're going to change `build:jsx` so it builds into a different directory (e.g., `build/`). Then after it's built files there, the simple thing would be to copy all the files at once, but the slightly better thing would be to create a script that copies the files only if they content has changed. `bin/script-utils.js` has an example of writing a file only if the content has changed, and that can serve as an example for copying.

IkkyOdufade:
Hi Ian, I've been trying to create a pr with just the npm build corrections but I cant. The deezer files keep appearing. I tried the git checkout origin/branch -- file but it kept saying error: pathspec 'deezer.js' did not match any file(s) known to git. Please help. Thank you.

IkkyOdufade:
Hi @ianb I would like to tell you how I have gone with this project since my pr is still not working properly. I changed the npm run build:js command, to create a new folder inside build, a lib folder. But the files being created are the ones inside eslintignore file. Is that how it is supposed to work?

ianb:
@IkkyOdufade If things get really messy, sometimes I just check out a new copy of the repository, or do `git reset --hard origin/master` (after you copy all your changed files someplace safe!)

What have to changed `build:jsx` to be? It should probably be something like `babel --relative --out-dir build/ --ignore gh-pages '**/*.jsx'` and then there should be files like `build/extension/options/optionsView.js`

IkkyOdufade:
[...]

PR #1346, opened 2020-03-19
npm run build:jsx

IkkyOdufade:
1. edited the package.json


IkkyOdufade:
Hi @ianb . Kindly find the edits I made. Looking forward to your corrections. Thank you.

ianb:
@IkkyOdufade Is this for #1193 ?

This will build the files into an alternate location, but it doesn't copy the results into their final destination. The files should be moved from `build/lib/...` into, for instance `extension/popup/popupController.js` (but only if the contents of the file changed). You'll probably have to write an additional script in `bin/` to move or copy the files.

IkkyOdufade:
Hi @ianb . So I tweaked the npm run build command again to `"build:jsx": "babel --relative --out-dir build/ --ignore gh-pages '**/*.jsx'",` to see what compiles and where it compiles to, and it compiles into 10 folders in the extension directory. I'm still trying to figure out how to write the script for checking and moving the files. The files compiled include `onboarding\onboardingController.js` and `onboarding\onboardingView.js`, in different folders.
npmbuild

ianb:
@IkkyOdufade After a quick google I came upon this: https://github.com/babel/babel/issues/8193

Apparently babel just doesn't work right :(

With a little experimentation, this _does_ work right: 

```sh
./node_modules/.bin/babel --relative  --out-dir ../build/ --only '**/*.jsx' extension/ --verbose
```

The weird `../build/` is because it's relative to the input directory `extension/`

IkkyOdufade:
Hi @ianb  . So is it only the changed files that gets moved to the build folder? Thank you for your help, I have been struggling with this.

ianb:
@IkkyOdufade Everything will get built by babel, but after that only changed files should be copied into their final location (alongside the `.jsx` file)

IkkyOdufade:
Oh ok. Do I still need to write the script that checks for changes?

IkkyOdufade:
@ianb , this code I tweaked kinda works.
`"build:jsx": "./node_modules/.bin/babel --relative  --out-dir build/ --only . --verbose '**/*.jsx' ",`
 So firstly it creates a build folder in the extension folder and the changed files are then copied into their original location in a new build folder. So like `onboarding\build\onboardingViews.js`.

PR #1347, opened 2020-03-19
Hides wakeword settings from production build

### LINKED ISSUE #1344: Hide wakeword pref in prod
The wakeword functionality shouldn't be exposed to prod builds, so we should just disable the pref for builds when `buildSettings.inDevelopment` isn't on.

It's good enough to just hide the pref on the preferences screen.

GangaChatrvedi:
Fixes #1344 

- setting _inDevelopment in (temporaryInstallId function) for cases where onInstalled do not get fired

### DIFF
--- a/extension/background/main.js
+++ b/extension/background/main.js
@@ -128,6 +128,8 @@ const temporaryInstallId = setTimeout(() => {
   if (_extensionTemporaryInstall === undefined) {
     _extensionTemporaryInstall = false;
   }
+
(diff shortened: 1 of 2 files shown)


GangaChatrvedi:
@ianb @JM-Mendez please take a look if this is fine.

ianb:
Ah, thank you, I had added buildSettings.inDevelopment recently but hadn't used it to update the other variables consistently.

Issue #1299, opened 2020-03-16
[Bang search] Linguee


awallin:
Based on a user feedback they would like to be able to Linguee by saying things like "search 'fox' with linguee."

There is a bang search for Linguee so it should be a straight forward implementation.
https://duckduckgo.com/bang?q=Linguee#

There's some more advanced implementations we could consider such as asking "say 'fox' on linguee" which would auto play the first audio file

annlinros:
Hi @awallin  I would like to take up this issue..

IkkyOdufade:
Hi @awallin can I work on this?

naima-shk:
Hi @awallin  Can i take this one please ? :)

awallin:
@annlinros give this a shot. The first part should be easy. I'll be curious what you think about adding an intent to auto play the audio with a search on Linguee.

PR #1354, opened 2020-03-20
Search on linguee

annlinros:
#1299 
I tried to implement the 'say' functionality by listing linguee as a music service just like spotify,deezer etc.
The search works partly, but the audio doesn't work yet.

Other implemenation I thought of was, manipulating `ddgBangSearchUrl()` in searching.js, to check if the service is linguee and then utter the query.


annlinros:
@awallin  @ianb  Please review.

PR #1355, opened 2020-03-21
Save to Pocket

### LINKED ISSUE #915: Save to Pocket
Based on telemetry users are trying to save the active page to Pocket.

"save this to pocket"
"save page to pocket"
"Pocket this"

JanviMahajan14:
Fixes #915

@ianb Do let me know if I approached to this issue in a right way.

### DIFF
--- a/extension/intents/pocket/pocket.js
+++ b/extension/intents/pocket/pocket.js
@@ -3,28 +3,25 @@ import * as intentRunner from "../../background/intentRunner.js";
 import * as content from "../../background/content.js";
 
 intentRunner.registerIntent({
-    name: "pocket.open",
-    async run(context) {
-        const url = "https://app.getpocket.com/";
-        await context.createTab({ url });
-    }
+  name: "pocket.open",
+  async run(context) {
+    const url = "https://app.getpocket.com/";
+    await context.createTab({ url });
+  },
 });
 
 intentRunner.registerIntent({
-    name: "pocket.save",
-    async run(context) {
-      const activeTab = await context.activeTab();
-      const metadata = await pageMetadata.getMetadata(activeTab.id);
-      await content.lazyInject(activeTab.id, [
-        "/background/pageMetadata-contentScript.js",
-        "/intents/pocket/pocketContentScript.js"
-      ]);
-      await browser.tabs.sendMessage(activeTab.id, {
-        type: "save",
-        url: metadata.url,
-      });
-    },
+  name: "pocket.save",
+  async run(context) {
+    const activeTab = await context.activeTab();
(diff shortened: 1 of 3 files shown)


extension/intents/pocket/pocket.js
```diff
@@ -0,0 +1,27 @@
+import * as pageMetadata from "../../background/pageMetadata.js";
+import * as intentRunner from "../../background/intentRunner.js";
+import * as content from "../../background/content.js";
+
+intentRunner.registerIntent({
+  name: "pocket.open",
+  async run(context) {
+    const url = "https://app.getpocket.com/";
+    await context.createTab({ url });
```
ianb: This could use `context.openOrFocusTab()` so it focuses an existing Pocket tab if it exists.

extension/intents/pocket/pocket.js
```diff
@@ -0,0 +1,27 @@
...
+    const activeTab = await context.activeTab();
+    const metadata = await pageMetadata.getMetadata(activeTab.id);
+    await content.lazyInject(activeTab.id, [
+      "/background/pageMetadata-contentScript.js",
+      "/intents/pocket/pocketContentScript.js",
+    ]);
+    await browser.tabs.sendMessage(activeTab.id, {
+      type: "save",
+      url: metadata.url,
+    });
```
ianb: This stuff doesn't actually have to happen inside the page itself, it can happen right here. That is, you can try to get the save URL right here.

extension/intents/pocket/pocketContentScript.js
```diff
@@ -0,0 +1,10 @@
+/* globals communicate */
+
+this.pocketContentScript = (function() {
+  communicate.register("save", async message => {
+    const url = message.url;
+    const xhr = new XMLHttpRequest();
+    xhr.open("GET", "https://getpocket.com/save?url=" + url, true);
+    xhr.send();
```
ianb: I'd prefer if this used [fetch](https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API) like:

```js
const resp = fetch(url);
if (resp.ok && !resp.redirected) {
  context.displayText("Saved");
}
```

Note also that quoting is important here, so use `"?url=" + encodeURIComponent(url)` – the current code will fail to save the query string of a page.

extension/intents/pocket/pocket.js
ianb: It might be even better to open the save URL in a hidden tab, and then close it if it goes OK, and focus it if not (e.g., when you aren't logged in).

ianb:
I think this is out of date now due to #1357

PR #1356, opened 2020-03-21
Reuse youtube tabs

### LINKED ISSUE #397: Reuse YouTube tab when it's the active tab.
We should open a new YouTube tab only when the active tab isn't a YouTube tab. Otherwise we'll reuse the tab.

amandhamija98:
This solves the following issue:
1. When a youtube tab is open and one says "play song name" the existing tab is used
Fixes #397

### DIFF
--- a/extension/background/intentRunner.js
+++ b/extension/background/intentRunner.js
@@ -153,8 +153,10 @@ export class IntentContext {
 
   async createTabGoogleLucky(query, options = {}) {
     const searchUrl = searching.googleSearchUrl(query, true);
-    const tab = !!options.youtubeTabId && options.youtubeTabId > -1? 
-    await browserUtil.loadUrl(options.youtubeTabId,searchUrl) : await this.createTab({ url: searchUrl });
+    const tab =
+      !!options.youtubeTabId && options.youtubeTabId > -1
+        ? await browserUtil.loadUrl(options.youtubeTabId, searchUrl)
+        : await this.createTab({ url: searchUrl });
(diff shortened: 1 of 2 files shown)


extension/background/intentRunner.js
```diff
@@ -146,7 +146,10 @@ export class IntentContext {
 
   async createTabGoogleLucky(query, options = {}) {
     const searchUrl = searching.googleSearchUrl(query, true);
-    const tab = await this.createTab({ url: searchUrl });
+    const tab =
+      !!options.youtubeTabId && options.youtubeTabId > -1
```
ianb: There's nothing YouTube-specific to this code, so you could just make this `openInTabId` instead of `youtubeTabId`

extension/services/youtube/youtube.js
```diff
@@ -23,8 +23,17 @@ async function waitForUrl(tabId, options) {
 }
 
 class YouTube extends serviceList.Service {
+  async getYoutubeTabId() {
+    const youtubeTabs = await this.getAllTabs({ currentWindow: true });
```
ianb: We don't want to reuse a background tab. I think this could probably be fixed just by adding `active: true` to the `getAllTabs` query.

amandhamija98:
Hi, I have added the code review comments, kindly take a look, thanks

ianb:
Looks good, thanks!

amandhamija98:
> Looks good, thanks!

Thank you!

PR #1357, opened 2020-03-21
Save to pocket functionality

annlinros:
#915 
Save to pocket works using the bookmarklet. 

 And Pocket extension gives feedback on
 - if the page was saved
 - if the user hadn't logged in.

Tests will fail because of the bookmarklet function( redeclaration of function arguments & var assignment).
I tried to manually revert errors (see  [contentScript.js]( https://github.com/mozilla/firefox-voice/pull/1357/commits/923caf2cba63e23208996a3616b1c871b2d0daff#diff-7510f54801fe7136e0c7e65168a87891 ) , 
then the bookmarklet won't work.

@ianb Kindly review.


annlinros:
Hi @ianb Can you review this PR when you get time ?

extension/intents/pocket/contentScript.js
```diff
@@ -0,0 +1,141 @@
+javascript: (function() {
```
ianb: I think this technically works with `javascript:` at the beginning because it looks like a label... but it should probably be removed.

Also we should include a comment about where the code came from (just giving the URL where it was found)

extension/intents/pocket/pocket.toml
```diff
@@ -0,0 +1,25 @@
+[pocket.open]
```
ianb: I don't see an implementation for `pocket.open` ?

annlinros:
@ianb sorry for missing out on the open intent..

annlinros:
@ianb I have made the changes per code review. Please have a look. Thank you.

PR #1360, opened 2020-03-21
Timer

### LINKED ISSUE #76: Native timer

danielamormocea:
Fixes #76

This PR implements the native timer.

- created a new directory for timer under intents/;
- added the intents for set, stop, reset, pause and unpause;
- created the class Timer that stores the timestamp (at which the timer has started or has been unpaused) and other details for the timer; This class also deals with notifying the pop-up when the timeout is done; 
- everytime the popup starts, it communicates with the timer in timer.js to get timestamp and the remaining time; (if the time > 0 it starts decreasing the timer locally)
- decided to use a timestamp to store and calculate the time, instead of using a shared object;

Modified a bit a functionality from popUpController to be able to reuse some functions; 
Added in utils a sleeper that also has the ability to clear the timeout. I needed this feature to stop the timer.

NOT YET:
This PR doesn't yet treat the follow-ups; (Waiting for the feature to be implemented in the other issue)
I am not verifying that the variables for hours, minutes, seconds are numbers in the toml (I can add them as a separate entity)
For the moment, intents like "set a timer for 2 minutes and a half", or "set a timer for 2 minutes and 2 seconds" are not supported. If this is necessary, I can add them later.

+ In intents, the close, pause, unpause and reset commands have a "timer" included to be able to differentiate; ("timer" can be discarded after the follow-ups)

Note: (The circle from the timer is bigger than in the mock-up because I tried to fit the hours as well).

@ianb  @awallin  Please take a look:) 

timer1
timer2
timer3


extension/popup/popupController.jsx
```diff
@@ -96,7 +132,8 @@ export const PopupController = function() {
     addListeners();
     updateExamples();
     updateLastIntent();
-    startRecorder();
+
+    if (skipRecording === false) startRecorder();
```
ianb: Hrm... I can understand the logic here, if the timer is running then don't start the recorder. But this means the whole tool will be disabled until the timer is done, right? The user doesn't have to say anything, so it feels like we should still have the recorder running even when there's a timer.

extension/popup/popupController.jsx
```diff
@@ -180,12 +217,59 @@ export const PopupController = function() {
...
   };
 
+  const clearTimer = async totalInSeconds => {
+    setPopupView("timer");
+
+    // use this variable to stop any other actions and show notifications
+    timerElapsed = true;
+
+    cancelRecoder();
+    playListeningChime();
```
ianb: We should make a followup to use a different bit of audio for the timer.

extension/popup/popupController.jsx
```diff
@@ -236,9 +326,19 @@ export const PopupController = function() {
   };
 
   const setPopupView = async newView => {
+    // do not change view if timer elapsed; timer has highest priority
+    if (timerElapsed === true) {
+      return;
+    }
```
ianb: This also doesn't seem right... but if we allow the recorder to run and treat the timer as an additional widget on the popup then I think this will be alright. That is, timer shouldn't be a "view", it'll just be an extra thing we show so long as the timer exists. Then there's still room for a card or anything else (the popup might get a little long, but that seems better than disabling functionality while the timer runs).

extension/intents/timer/timer.js
```diff
@@ -0,0 +1,186 @@
...
+      }
+    } catch (e) {
+      catcher.capture(e);
+    }
+
+    return browser.experiments.voice.openPopup();
+  }
+}
+
+export const timer = new Timer();
```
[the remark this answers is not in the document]
danielamormocea: @ianb I am not sure what you mean; Should I make a TimerController that has the methods getActiveTimer, closeActiveTimer, startActiveTimer...and use an instance of Timer to perform the logic of the timer and the TimerController to do the dispatching and to encapsulate the timer creation logic? And now we would have only one instance of TimerController; Or are you thinking of something else?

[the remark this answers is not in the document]
ianb: I guess it makes sense to me that if I saw "start timer for 30 seconds" and then the timer finishes, then I saw "start timer for 1 minute" the second timer should be a new object. You could use TimerController or just make getActiveTimer() an exported function. (`.close()` and `.start()` could be methods on that timer object - only getting the timer object has to be an exported function)

extension/popup/popupController.jsx
danielamormocea: @ianb the recorder works while the timer is active; I am not sure what should be changed right now; I will push the changes requested shortly and let me know if the timer doesn't work right.

danielamormocea:
@ianb I made the requested changes. Please take a look when you have time :) Also, I am not sure why the timer did not work for you last time; from what you described it seemed like the jsx was not built (all the changes I made in the popup view and controller were not seen, but the changes in the timer.js were) ?! Please let me know if it happens again

extension/intents/timer/timer.js
```diff
@@ -0,0 +1,183 @@
...
+    if (this.timeoutId !== undefined) clearTimeout(this.timeoutId);
+
+    return this.start(this.totalInMS);
+  }
+
+  async start(duration = this.totalInMS) {
+    this.startTimestamp = new Date().getTime();
+    this.remainingInMS = duration;
+
+    this.timeoutId = setTimeout(() => this.openPopup(), duration);
```
ianb: There's a few blank lines and I prefer keeping braces around all conditionals, but not blockers...

PR #1362, opened 2020-03-23
Fixes #1256 - config: integrate selenium in ci

### LINKED ISSUE #1256: Enable Selenium test in CI

Simpcyclassy:


Simpcyclassy:
@ianb please could you look at my PR?

ianb:
I don't think this will actually run the Selenium tests or start Firefox. I'm not exactly sure what the Selenium server is, but I don't think it's necessary for what we're doing, we'll be using Firefox (via a package called geckodriver) and the test script we have should start and connect to Firefox.

Simpcyclassy:
Alright that makes sense

ianb:
I'm guessing this is stalled, please reopen or recreate if you want to add to it.

PR #1371, opened 2020-03-24
Example fixes

farhatcode:
I happened to come across `tabs.previous` intent in tabs.toml, that match could be improved. Then gave a quick glance to all other intents and come up with another example correction.

### DIFF
--- a/extension/intents/tabs/tabs.toml
+++ b/extension/intents/tabs/tabs.toml
@@ -205,7 +205,7 @@ phrase = "Reset zoom"
 [tabs.previous]
 description = "Go to a previously active tab"
 match = """
-  (Go | switch | change | activate |) (the | my |) previous tab
+  (Go | switch | change | activate |) (to |) (the | my |) previous tab
 """
 
 [[tabs.previous.example]]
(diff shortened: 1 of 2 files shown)


extension/intents/tabs/tabs.toml
```diff
@@ -205,7 +205,7 @@ phrase = "Reset zoom"
 [tabs.previous]
 description = "Go to a previously active tab"
 match = """
-  (Go | switch | change | activate |) (the | my |) previous tab
+  (Go | switch | change | activate |) (to |) (the | my |) previous tab
```
ianb: That would match "go to my previous tab" or "switch to my previous tab" etc. So I think we'd want to keep it.

farhatcode: I have updated it @ianb.

farhatcode: or may I close this PR because this change was not required?

ianb:
@farhatcode Yes, it occurs to me that the testing example probably didn't need updating – I think it was something that I actually encountered as a mistranscription and wanted to be sure it worked. There's even some special code for repeated words: https://github.com/mozilla/firefox-voice/blob/master/extension/background/language/textMatching.js#L94-L100

farhatcode:
Thanks @ian for explaining.

PR #1382, opened 2020-03-26
Fixes #916 restart firefox

### LINKED ISSUE #916: Restart Firefox
May not be possible but user has tried to use "restart Firefox" restart the browser.

Simpcyclassy:
I  was wondering if there is an existing API for restart @ianb. Cause it seems to me the browser API to shut down firefox makes Firefox itself kill the processes of all extensions.

I am not sure if there's a way to do this outside of child processes. There seems to be no way to do this outside Firefox exposing a browser restart application command to extensions

### DIFF
--- a/extension/experiments/voice/api.js
+++ b/extension/experiments/voice/api.js
@@ -54,6 +54,10 @@ this.voice = class extends ExtensionAPI {
           async quitApplication() {
             return runCommand("cmd_quitApplication");
           },
+
+          async restorePreviousSession() {
+            return runCommand("Browser:RestoreLastSession");
+          }
         },
       },
     };
--- a/extension/experiments/voice/api.js
+++ b/extension/experiments/voice/api.js
@@ -43,8 +43,16 @@ this.voice = class extends ExtensionAPI {
             return runCommand("History:UndoCloseTab");
           },
 
+          async closeWindow() {
+            return runCommand("cmd_closeWindow");
+          },
+
           async undoCloseWindow() {
             return runCommand("History:UndoCloseWindow");
+          },
+
+          async OpenBrowserWindow() {
+            return runCommand("cmd_newNavigator");
           },
 
           async openDownloads() {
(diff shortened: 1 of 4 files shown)


Simpcyclassy:
Is there an existing api for restarting firefox @ ianb

ianb:
@Simpcyclassy you are on the right track by updating api.js. This would only work with privileged code. Unfortunately not every command is clearly exposed, or some commands don't use the same technique. `Browser:RestoreLastSession` seems to close the browser without restarting, for instance.

This seems to be the code that restarts the browser in the About dialog: https://searchfox.org/mozilla-central/source/browser/base/content/aboutDialog-appUpdater-legacy.js#245-277 – but I don't know how you'd call the `.buttonRestartAfterDownload()` method.

Simpcyclassy:
So, I do notice in other parts of the code, when an action is to be performed may be like something one can do from the settings pages, the HTML is loaded and the element needed to perform the action is  "clicked". I think  button for restarting the browser probably only shows up under certain specific conditions. But then this looks doable, I will work my way around it.

Simpcyclassy:
It's kinda tricky knowing which browser API to call, so I am trying to take an approach where I close the window, reopen and restore the previous session but this seems to work for me only when no tabs are open.
Then for the close window intent runner on line 48, I was wondering if there was a specific need to get all windows before closing or `cmd_closeWindow` is enough to handle all that?

ianb:
I don't think any combination of closing and opening windows or tabs will work for this: I believe the reason someone would want to run "restart Firefox" is they want to apply the Firefox update, or Firefox is causing some issue and they want to restart it to fix it. So the entire process has to entirely quick and restart. This is something Firefox can do, as there's the button that does this, but it's not restarting _part_ of Firefox, it's restarting everything.

Simpcyclassy:
I will look more into this and see how  I can call the `.buttonRestartAfterDownload()` method

ianb:
I don't think we have a solution for this, so I'll close this for now. If you continue with it, please reopen.

Simpcyclassy:
No problem

PR #1403, opened 2020-03-28
Fix tabs not loading properly when using routines

### LINKED ISSUE #1351: New routine doesn't complete all actions when made a second time
### Prerequisites:
A fresh new Firefox profile is made and the extension is installed for the first time on the profile.
Mic permissions are enabled.
A routine is made that contains more than 1 action. (click mic icon next to the window's menu button, click the "Settings" button from the doorhanger, click "Routines" from the left side of the "Preferences" page, click "+ New routine" button, fill out the fields with the routine name and the actions that need to be done and click the "Save" button) 
Example of routine:
Annotation 2020-03-20 154903

### STR:
3. Repeat steps 1 and 2 after closing the new tabs that opened.

### Expected result:
New York Times, CNN and Vox are opened again in new tabs.

### Actual result:
Only Vox (the last action from the routine) opens in a new tab.

### Notes:
Reproduced on Mac 10.14 and Win10x64 with Firefox Nightly 76.0a1 (2020-03-20).
It seems when the "News" command is made a second time, it wants to start to open the New York Times tab, but it just jumps to open Vox instead.
Seems to apply for other new routines made in the new Firefox profile, but it may depend on the routine actions that need to be completed . ("books" routine - 2nd gif)

### LINKED ISSUE #1364: Last routine action opens if there are more "Search Google Images" commands
### Prerequisites:
Mic permissions are enabled.
A routine is made that contains more than 1 "Search Google Images" command. (click mic icon next to the window's menu button, click the "Settings" button from the doorhanger, click "Routines" from the left side of the "Preferences" page, click "+ New routine" button, fill out the fields with the routine name and the actions that need to be done and click the "Save" button)

### STR:
2. Say or write the newly made routine command, e.g. "Dogs".

### Expected result:

### Actual result:
Only one Google Images tab opens corresponding to the last routine action.

### Notes:

### LINKED ISSUE #1365: "Open Google" blocks other actions if placed first in a routine
### Prerequisites:
Mic permissions are enabled.

### STR:

### Expected result:

### Actual result:

### Notes:

dave-ok:

Fixes #1365 . Possibly Fixes #1364 and Fixes #1351
@ianb Could you please make out time to review this? The function `createTab` in `browserUtils.js`, decides to reuse an existing tab partly on if the active tab's `url` is `about:blank`. It seems the browser does not immediately resolve the `url` when a new tab is loading, even after the `tab.status` changes to `complete` and the tab is supposedly loaded, it still resolves to `about:blank`. Included a condition to ensure the page is not busy loading and if loaded that it actually has a title. This was the only way I found to test if the page is actually a blank page. It seems Chrome has a `pendingUrl` property in tabs for this reason, but I couldn't get something similar for Firefox. Thanks

### DIFF
--- a/extension/browserUtil.js
+++ b/extension/browserUtil.js
@@ -83,8 +83,8 @@ export async function activateTabClickHandler(event) {
 
 export async function createTab(options = {}) {
   const active = await activeTab();
-  if (
-    ["about:blank", "about:home", "about:newtab"].includes(active.url) ||
+  if ((["about:blank", "about:home", "about:newtab"].includes(active.url) &&
+    !(active.status === "loading") && (active.title === "")) ||
     (buildSettings.executeIntentUrl &&
       active.url.startsWith(buildSettings.executeIntentUrl))
   ) {
--- a/extension/browserUtil.js
+++ b/extension/browserUtil.js
@@ -83,8 +83,10 @@ export async function activateTabClickHandler(event) {
 
 export async function createTab(options = {}) {
   const active = await activeTab();
-  if ((["about:blank", "about:home", "about:newtab"].includes(active.url) &&
-    !(active.status === "loading") && (active.title === "")) ||
+  if (
+    (["about:blank", "about:home", "about:newtab"].includes(active.url) &&
+      !(active.status === "loading") &&
+      active.title === "") ||
     (buildSettings.executeIntentUrl &&
       active.url.startsWith(buildSettings.executeIntentUrl))
   ) {


ianb:
@dave-ok This seems reasonable as a workaround.

Probably the right fix in these cases is to fix the intent that led up to this problem. For instance, if you say "go to new york times" then that intent should only finish when the nytimes.com site actually loads (I'm not sure if it does or not). If the intent returns before the page loads, then the next intent in a routine could fail. I think your fix could help in that case, but the ideal fix is for the first intent not to finish until it is really finished.

A common fix for this is to use `browserUtil.loadUrl()`, which actually checks that the tab loads something. But I added that relatively recently, and I'm sure it isn't being used everywhere it should be (and maybe we need other forms of that same approach).

dave-ok:
@ianb Thank you. I did not know about the `browserUtil.loadUrl()` function. I am still trying to figure out the purpose and area of use of some of these utility functions. But, after making this PR while going through the codebase I noticed another utility function `browserUtil.waitForDocumentToComplete`. Does this have a similar functionality?

PR #1376, opened 2020-03-25
Fix positioning of text when using "Dictate"

dave-ok:
Before submitting a final PR, please:

- [ ] Run `npm test` on your machine
- [x] Run `npm run format` to keep code formatted consistently
- [x] Reference the issue you are fixing in at least one of your commit messages, as `Fixes #X`

This PR fixes issue #1373. @ianb Could you please review this when you can? Thanks

PR #1405, opened 2020-03-28
Fix rich text editor cursor positioning

### LINKED ISSUE #1373: Another "Dictate" command can't be done before the initial "Dictate" command transcription
### Prerequisites:
Mic permissions are enabled.
An online notepad tab is opened and made active, e.g. [calmlywriter](https://www.calmlywriter.com/online/) or [anotepad](https://anotepad.com/).

### STR:
2. Say or write a "Dictate" command, e.g. "Type don't forget to check emails".
3. Move the cursor before the transcribed text and pres "Enter" a few times.
4. Move cursor back to the top and repeat steps 1 and 2.

### Expected result:
The text is transcribed successfully in the tab.

### Actual result:
New text is added after the initial transcription.

### Notes:
Reproduced on Mac 10.14 and Win10x64 with Firefox Nightly 76.0a1 (2020-03-24).
Not reproduced if cursor is moved anywhere after the initial transcription.

dave-ok:

Fixes #1373 
This PR handles string insertion for rich text editors when dictating. This is a supplement to #1376.
@ianb apologies for the oversight in previous PR. Could you review this when you can?

### DIFF
--- a/extension/intents/forms/formsContentScript.js
+++ b/extension/intents/forms/formsContentScript.js
@@ -89,11 +89,7 @@ this.dictationContentScript = (function() {
     setTimeout(() => {
       if (el.hasAttribute("contenteditable")) {
         // eslint-disable-next-line no-unsanitized/property
-        el.innerHTML = insertString(
-          el.innerHTML,
-          quoteHtml(text),
-          el.selectionStart
-        );
+        insertStringContentEditable(quoteHtml(text), el);
       } else {
         el.value = insertString(el.value, text, el.selectionStart);
       }
@@ -246,6 +242,51 @@ this.dictationContentScript = (function() {
     }
 
     return `${startSlice}${spaceBefore}${snippet}${spaceAfter}${endSlice}`;
+  }
+
+  function insertStringContentEditable(snippet, containerElement) {
+    if (window.getSelection) {
+        const sel = window.getSelection();
+        if (sel.getRangeAt && sel.rangeCount) {
+            let range = sel.getRangeAt(0);
+            // range.deleteContents(); TODO Maybe: overwrite selection
+
+            const modSnippet = prependAppendSpace(snippet, containerElement, range);
+
+            const snippetNode = document.createTextNode(modSnippet);
+            range.insertNode(snippetNode);
+
+            range = range.cloneRange();
+            range.setStartAfter(snippetNode);
+            range.collapse(true);
+            sel.removeAllRanges();
+            sel.addRange(range);
+        }
+    }
+  }
+
+  function prependAppendSpace(snippet, containerElement, range) {
+    let strAfter, strBefore;
+    const preRange = range.cloneRange();
+    const postRange = range.cloneRange();
+    preRange.collapse(true);
+    preRange.setStart(containerElement, 0);
+    if (preRange.toString().slice(-1) === "" || preRange.toString().slice(-1) === " ") {
+      strBefore = "";
+    } else {
+      strBefore = " ";
+    }
+
+    postRange.collapse(false);
+    postRange.setEndAfter(containerElement);
+
+    if (postRange.toString().startsWith(" ")) {
+      strAfter = "";
+    } else {
+      strAfter = " ";
+    }
+
+    return `${strBefore}${snippet}${strAfter}`;
   }
 
   function focusDirection(dir) {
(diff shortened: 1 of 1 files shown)


extension/intents/forms/formsContentScript.js
```diff
@@ -89,11 +89,7 @@ this.dictationContentScript = (function() {
     setTimeout(() => {
       if (el.hasAttribute("contenteditable")) {
         // eslint-disable-next-line no-unsanitized/property
-        el.innerHTML = insertString(
-          el.innerHTML,
-          quoteHtml(text),
-          el.selectionStart
-        );
+        insertStringContentEditable(quoteHtml(text), el);
```
ianb: Generally quoting should happen right where the text is used, then someone can read insertion code and know that things are quoted properly. But in this case no quoting is necessary, since `createTextNode` will do the quoting. Probably if you try to insert `"` with this code you'll get `&quot;`

extension/intents/forms/formsContentScript.js
```diff
@@ -89,11 +89,7 @@ this.dictationContentScript = (function() {
     setTimeout(() => {
       if (el.hasAttribute("contenteditable")) {
         // eslint-disable-next-line no-unsanitized/property
```
ianb: This comment can be removed now, since it doesn't use innerHTML

extension/intents/forms/formsContentScript.js
```diff
@@ -248,6 +244,51 @@ this.dictationContentScript = (function() {
     return `${startSlice}${spaceBefore}${snippet}${spaceAfter}${endSlice}`;
   }
 
+  function insertStringContentEditable(snippet, containerElement) {
+    if (window.getSelection) {
```
ianb: This function looks like it needs reformatting.

You don't need to test `window.getSelection` – we know this code is running in a modern Firefox browser. Also, if we can detect it doesn't work, we should just raise an exception. This code doesn't do anything, which will be confusing to users. But here I'd just remove the tests for `window.getSelection` and `sel.getRangeAt`

extension/intents/forms/formsContentScript.js
```diff
@@ -248,6 +244,51 @@ this.dictationContentScript = (function() {
...
+        const sel = window.getSelection();
+        if (sel.getRangeAt && sel.rangeCount) {
```
ianb: I think `sel.rangeCount` could be zero though, meaning there's no selection. That should probably default to appending then.

extension/intents/forms/formsContentScript.js
```diff
@@ -248,6 +244,51 @@ this.dictationContentScript = (function() {
...
+    }
+  }
+
+  function prependAppendSpace(snippet, containerElement, range) {
+    let strAfter, strBefore;
+    const preRange = range.cloneRange();
+    const postRange = range.cloneRange();
+    preRange.collapse(true);
+    preRange.setStart(containerElement, 0);
+    if (preRange.toString().slice(-1) === "" || preRange.toString().slice(-1) === " ") {
```
ianb: can you use `preRange.toString().endsWith(" ")` here?

dave-ok:
Thank you for these recommendations I have already started work on them

extension/intents/forms/formsContentScript.js
dave-ok: I did not realize that. I have made the changes. And I had to remove the function because it was no longer being used

extension/intents/forms/formsContentScript.js
dave-ok: Done

extension/intents/forms/formsContentScript.js
dave-ok: Redundant tests everywhere! Sorry about that :) I have reformatted the function

extension/intents/forms/formsContentScript.js
dave-ok: Turns out `sel.rangeCount` is not needed as well as it is meant for multiple ranges/selections. The function only needs the cursor position which is provided by `sel.getRangeAt(0)`

extension/intents/forms/formsContentScript.js
dave-ok: Done

dave-ok:
@ianb I have made the changes you requested. Could please review them?

PR #1406, opened 2020-03-29
At least 3 bookmarks with similar/same title can all open

Simpcyclassy:


Simpcyclassy:
Could you look at my PR @ianb

ianb:
@awallin what are your thoughts on this? This will open multiple tabs if there are multiple bookmarks that match the query

awallin:
Not sure I understand what's happening? Are multiple bookmarks opening in new tabs if there are similar matches? If that's the case I'd say we should make the best decision for the user and open the closest match. (Until we implement some conversational/confirmation feature)

Simpcyclassy:
So the current implementation does not necessarily open the best match. It looks for all bookmark that matches search and opens the one that was saved last which happens to be the first item in the array. 
With the PR I raised it opens at least the first 3 items in the array if there are so many matches, maybe the user might be looking for a bookmark that was not the last saved.

awallin:
When making a decision about the behavior in a case like this I like to fall back on our design principles for Firefox Voice. One of them is "It just works. It makes good choices for me." 

Ideally we'd open the bookmark that was the best match. I worry that opening multiple matches is a poor experience for the user. They likely only want one and we have a chance to satisfy that by applying some matching logic but opening all 3 will always result in extra work for the user as they will have to close the tabs they didn't intent to open.

While simply relying recency probably isn't the best matching pattern (although it's a relevant factor) I'd suggest thinking about ways to improve that vs opening multiple matches.

Simpcyclassy:
Makes a lot of sense to me. I will look into that

PR #1412, opened 2020-03-30
Npm run build:jsx

IkkyOdufade:
1. Edited the npm run build:jsx
2. Creates a build folder in the original location of the files
3. Creates the .js files in the original location of the jas files.


IkkyOdufade:
Hi, @ianb  I edited the command and was able to build js files in the original location. I'm sorry its taking so long to finish. Pls help with corrections.

PR #1413, opened 2020-03-30
fixes #388[browser command] Open Settings / Preferences

### LINKED ISSUE #388: [browser command] Open Settings / Preferences
Based on telemetry users want to open the browsers setting and preferences. Implement the following:

"Open settings" or "open preferences" = Load a new about:preferences tab
"Open about debug" or "open debugging" = Load a new about:debugging tab
"Open extensions" or "open add-ons" = Load a new about:addons tab

Simpcyclassy:
I do not seem to find any command for about:debugging


Simpcyclassy:
Please have a look at my PR @ianb

extension/experiments/voice/api.js
```diff
@@ -54,6 +54,15 @@ this.voice = class extends ExtensionAPI {
           async quitApplication() {
             return runCommand("cmd_quitApplication");
           },
+
+          async openPreferences() {
+            return runCommand("Browser:OpenAboutContainers");
+          },
+
+          async browserOpenAddonsMgr() {
+            return runCommand("Tools:Addons");
```
ianb, line 63: I kind of wish we had a general way to open `about:` URLs. I've done that before, but I'm a little afraid that the code is fragile, or might even be broken right now: https://github.com/ianb/sticky-reader-mode/blob/master/addon/experiment/api.js

What you've done is probably safer for now.

extension/experiments/voice/api.js
```diff
@@ -54,6 +54,15 @@ this.voice = class extends ExtensionAPI {
           async quitApplication() {
             return runCommand("cmd_quitApplication");
           },
+
+          async openPreferences() {
+            return runCommand("Browser:OpenAboutContainers");
```
ianb, line 59: This goes to the same thing as "Manage Containers", which is technically in preferences, but not a great solution. 

It works for me if this calls `return runCommand("menu_preferences");` which points to a menuitem (but works just as well as a `<command>`): https://searchfox.org/mozilla-central/source/browser/base/content/browser-menubar.inc#136

extension/intents/browser/browser.toml
```diff
@@ -0,0 +1,37 @@
+[browser.preferences]
+description = "open browser settings or preferences"
+match = """
+  open browser (settings | preferences |)
```
ianb, line 4: `(browser | firefox)`

Simpcyclassy: I just added the alternative match

extension/experiments/voice/api.js
Simpcyclassy: OK cool. It will come in very handy when fixing. about:debugging

extension/experiments/voice/api.js
Simpcyclassy: I have this fixed. So, if I wanted to use this as `<command>`, how will I implement it? What will it look like in the code?

ianb:
@Simpcyclassy it looks like you lost your `api.js` changes with the new commits. It looks like you had the right code in a commit (31acbb5b771fd6bd6993c6ef30d5a126f13ed8f3) but something after it may have reverted it

Simpcyclassy:
Thanks I have fixed it now

PR #1416, opened 2020-03-30
search and say functionality for Linguee

annlinros:
Hi @ianb 
Could you please review this PR for  #1299 ?

I tried to access the element with the audio class on it using different ways.But none of that catches the element.
Except that bang search works.


awallin:
Tested a works great for me.

annlinros:
@awallin  does the "say fox on linguee" work ? Beacause I was having trouble making it work.

awallin:
@annlinros "say fox on Linguee" returns the result in Linguee but doesn't play the audio.

Can you also enable "search fox on Linguee?" Currently that returns a Google result.

annlinros:
Hi @awallin  I have created a new PR for this #1462 , could you please review that ? Thanks.
I kinda accidently messed up the git history.

awallin:
@annlinros for me `search` and `say` both return the correct result on Linguee but `say` doesn't play the audio.

PR #1293, opened 2020-03-16
fixes #353 - added support for spotify album/playlist intent

melvin2016:
Before submitting a final PR, please:

- [x] Run `npm test` on your machine
- [x] Run `npm run format` to keep code formatted consistently
- [x] Reference the issue you are fixing in at least one of your commit messages, as `Fixes #X`

Fixes #353 .

PR #1422, opened 2020-03-30
Followup to #1293, remove irrelevant minCount: 0

ianb:


extension/services/spotify/player.js
```diff
@@ -30,8 +30,6 @@ this.player = (function() {
         try {
           const playerButton = await this.waitForSelector(SEARCH_PLAY, {
             timeout: 10000,
-            // There is no need to wait for this as there is only one selector
-            minCount: 0,
           });
```
Simpcyclassy: What was the initial idea behind adding minCount?

ianb: It used to be `minCount: 4`, which meant only return when there were at least 4 elements that matched the selector. (For reasons I never looked into closely, Spotify had several buttons that matched the previous selector but didn't all work.)

Simpcyclassy: Ohh. I see

PR #1426, opened 2020-03-30
[browser command] bookmark all tabs

Simpcyclassy:

### DIFF
--- a/extension/experiments/voice/api.js
+++ b/extension/experiments/voice/api.js
@@ -67,6 +67,10 @@ this.voice = class extends ExtensionAPI {
             return runCommand("Browser:ShowAllHistory");
           },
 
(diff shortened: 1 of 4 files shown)


Simpcyclassy:
Please could you look my PR @ianb

ianb:
I don't think this particular function makes sense. It leads to a dialog that requires the keyboard, and the functionality itself is somewhat obscure:

image

We could do this just through the bookmarks API (and then the name could be part of the utterance), but I don't think bookmarking all tabs is necessary, unless @awallin feels otherwise.

awallin:
Yea, not sure I see a realistic scenario where someone wants to bookmark every tab they have open.

ianb:
I think @awallin and I agree that this isn't something people will want to use.

PR #1432, opened 2020-03-31
Use chrono for timer

### LINKED ISSUE #1379: [Timer] Use Chrono
Followup to #1360:

Currently it gets the numbers explicitly from the phrases, but we could use [Chrono](https://github.com/wanasit/chrono) to parse times. It seems to handle many of the cases appropriate for this: time in the future, minutes, minutes and seconds, etc.

In the phrase we'd just capture an untyped `[time]` slot, and then use Chrono to parse that time, instead of having separate hours/minutes/seconds.

danielamormocea:
Fixes #1379

This PR uses chrono for parsing time in timer. Now, the timer works with phrases like "set timer for 2 hours and 30 minutes".

@ianb Check this when you have time, please :)
We could also implement the alarm feature with this (similar with timer, but different behaviour for pause / unpause).


danielamormocea:
@ianb  I added the error check, and also I noticed that when the timer is set for "0 minutes/seconds/hours" the chrono behaves a bit weird (calculates the time until 00:00 and then timer starts with that countdown), so I fixed this problem also :)

PR #1442, opened 2020-03-31
Npm/build

IkkyOdufade:
1. edited the package.json to allow changed files to be built into their original locations.


package.json
```diff
@@ -95,7 +95,7 @@
     "build:manifest": "node bin/substitute-manifest.js",
     "build:intent-metadata": "node bin/parse-intent-toml.js",
     "build:ppn-listing": "node bin/generate-ppn-listing.js > extension/js/vendor/ppnListing.js",
-    "build:jsx": "babel --relative --out-dir . --ignore gh-pages '**/*.jsx'",
+    "build:jsx": "./node_modules/.bin/babel --relative  --out-dir ./ --only '**/*.jsx' extension/ --verbose",
```
ianb, line 98: In package.json the `$PATH` is updated to include `./node_modules/.bin` so you don't need the full path here. The full path is for use on the command line when you want to test things out. Similarly `--verbose` is for testing.

IkkyOdufade:
Hi, @ianb . I've made the changes you requested for. Waiting for your feedback.

bin/npmbuild-script.js
```diff
@@ -0,0 +1,38 @@
+/* eslint-disable no-console */
+const fs = require("fs");
+const DIR = "./build";
+
+fs.readdir(DIR, (err, files) => {
```
ianb, line 5: For most of these `fs` routines there's a `*Sync` version. So `fs.readdirSync`. These don't require callbacks. It's not something you should use in a server (it will block the process), but for scripts like this it's perfect.

IkkyOdufade: Hi, @ianb. Thank you for this. Woking on this issue really taught me a lot. I know it is supposed to be easy, but for me, it took me 3 weeks.

PR #1444, opened 2020-03-31
edited config.yml to include jest

IkkyOdufade:
1. edited config.yml to include jest. 

Hi, @ianb. The tests passed, but I know this is not solving the issue, so can I get more guidance? 
image


IkkyOdufade:
Hi, @ianb . After I changed babelrc to babel.config.json and added @babel/preset-react, one test related to the integration.test.js keeps failing. 
image

ianb:
@IkkyOdufade That test seems to be failing locally as well. It looks like we need to run `npm run package` before the test runs, and that will create `./addon.xpi`. Then in `integration.test.js` update `addonFileLocation` to find `addon.xpi`

PR #1445, opened 2020-03-31
[Alias] bookmarks and history intents

Simpcyclassy:
@awallin here are the aliases as requested.

@ianb please could you have a look


ianb:
These look good, thanks!

Simpcyclassy:
You're welcome!

PR #1446, opened 2020-04-01
Fix google navigation blocking other actions

### LINKED ISSUE #1365: "Open Google" blocks other actions if placed first in a routine

dave-ok:

Fixes #1365
 
@ianb I tried but failed to understand the reason for testing for [google domain](https://github.com/mozilla/firefox-voice/blob/5078ef0840e96e9fe36e2800dbada3f0bf7b1039/extension/background/intentRunner.js#L174) in the `onUpdated` function within the Promise in `createTabGoogleLucky`. Either way it turned out that if a `google.com` url is being loaded the promise never resolves. Introducing an else condition seems to solve this. Am I on the right track?

The utterance `open google` searches for 'open' on google through the `search.searchGoogle` intent instead of using the `navigate.navigate` intent. The matcher in its current state matches phrases like "_Search for cars google_", "_Look for cute kittens google for me_", "_open google_" etc. Modified it to match only when "_on google_" is present. So "_Search for cars on google_" and "_Look for cute kittens on google for me_" will match whereas "_open google_" will not.

Could you please review this? Thanks

### DIFF
--- a/extension/background/intentRunner.js
+++ b/extension/background/intentRunner.js
@@ -189,6 +189,10 @@ export class IntentContext {
           // We no longer need to listen for updates:
           browserUtil.onUpdatedRemove(onUpdated, tab.id);
           resolve(tab);
+        } else {
+          // We no longer need to listen for updates:
+          browserUtil.onUpdatedRemove(onUpdated, tab.id);
+          resolve(tab);
         }
       }
       try {
--- a/extension/background/intentRunner.js
+++ b/extension/background/intentRunner.js
@@ -171,7 +171,7 @@ export class IntentContext {
         ) {
           return;
         }
-        const isGoogle = /^https:\/\/www.google.com\//.test(url);
+        const isGoogle = /\/search/.test(url);
         const isRedirect = /^https:\/\/www.google.com\/url\?/.test(url);
         if (!isGoogle || isRedirect) {
           if (isRedirect) {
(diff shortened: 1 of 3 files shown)


ianb:
@dave-ok I think this is probably not the correct result.

An "I'm lucky" search goes to `https://www.google.com/search?q=query&btnI=` and then often (but not always) goes to `https://www.google.com/url?q=newUrl` and then to the final URL.

I think the issue here is that sometimes the final URL is _also_ a google.com URL, and then it never "finishes" because finishing is defined as leaving google.com.

I think simply improving the `isGoogle` test to match only `/search` might fix this?

dave-ok:
Thanks. That makes things so much clearer. I'll make the changes

dave-ok:
@ianb I have modified the test. Can you please check it?

extension/background/intentRunner.js
```diff
@@ -171,7 +171,7 @@ export class IntentContext {
         ) {
           return;
         }
-        const isGoogle = /^https:\/\/www.google.com\//.test(url);
+        const isGoogle = /\/search/.test(url);
```
ianb, line 174: I think a somewhat more careful test here might be in order, but I might put it in real quick after landing this

dave-ok:
@ianb I guess I misunderstood the point you were trying to pass across. Thank you very much

PR #1451, opened 2020-04-01
Fix scrolling not working on some sites

### LINKED ISSUE #1433: Scrolling not working on Wikipedia and other pages
### Prerequisites:
Mic permissions are enabled.
"A privacy reminder from Google" is reviewed and agreed and the yellow language pop-up is closed. (if applicable)
Have an active Wikipedia tab opened.

### STR:
2. Say or write a "Scroll" command, e.g. "Scroll down", "Scroll to top" (if page scroll bar is not at the top), "Scroll all the way down", "Scroll up".

### Expected result:
Page scrollbar acts accordingly to the command made.

### Actual result:
Nothing happens.

### Notes:
Reproduced on Mac 10.14 with Firefox Nightly 76.0a1 (2020-03-31).
Reproduced also on Instagram, Twitter.
Not reproduced on Youtube, Facebook, Bugzilla.

dave-ok:

Fixes #1433 

@ianb After some hair pulling, I abandoned the quest to look for a different solution. It turns out the existing solution handles scrolling very well, apart from a few tweaks needed, just as you initially suggested.

Some sites like wikipedia, instagram etc., set their <body> and <html> height to 100%. The scrolling is delegated to the <html>. So the test for scrollable element is bumped up to the parentNode, <html>, for these cases. When you get the chance can you please review this? Thanks

### DIFF
--- a/extension/intents/scroll/scrollHelper.js
+++ b/extension/intents/scroll/scrollHelper.js
@@ -8,8 +8,12 @@ function getScrollParent(node) {
   }
 
   if (node.scrollHeight > node.clientHeight) {
-    if (node.scrollHeight === node.parentNode.scrollHeight) {
-      return node.parentNode;
+    // if on recursion we get to the body element
+    // body element is not scrollable in firefox, return documentElement
+    // https://developer.mozilla.org/en-US/docs/Web/API/document/scrollingElement
+
+    if (node.tagName === "BODY") {
+      return document.documentElement;
     }
     return node;
   }
--- a/extension/intents/scroll/scrollHelper.js
+++ b/extension/intents/scroll/scrollHelper.js
@@ -12,7 +12,7 @@ function getScrollParent(node) {
     // body element is not scrollable in firefox, return documentElement
     // https://developer.mozilla.org/en-US/docs/Web/API/document/scrollingElement
 
-    if (node.tagName === "BODY") {
+    if (node === document.body) {
       return document.documentElement;
     }
     return node;
--- a/extension/intents/scroll/scrollHelper.js
+++ b/extension/intents/scroll/scrollHelper.js
@@ -8,6 +8,9 @@ function getScrollParent(node) {
   }
 
   if (node.scrollHeight > node.clientHeight) {
+    if (node.scrollHeight === node.parentNode.scrollHeight) {
+      return node.parentNode;
+    }
     return node;
   }
   return getScrollParent(node.parentNode);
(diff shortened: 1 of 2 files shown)


extension/intents/scroll/scrollHelper.js
```diff
@@ -8,6 +8,9 @@ function getScrollParent(node) {
   }
 
   if (node.scrollHeight > node.clientHeight) {
+    if (node.scrollHeight === node.parentNode.scrollHeight) {
+      return node.parentNode;
+    }
```
ianb, line 13: What does this do? I'm a little confused here... the first test sees if the node can be scrolled (node.scrollHeight > node.clientHeight), but the second test... sees if the entire height of the current node (including portions scrolled off) is equal to the entire height of the parent node (irregardless of the rendered size of the parent node) and then uses that...

Also I wonder, if this is the right test (even though I don't understand it) why just return the parentNode? Why not return the further ancestor for which this is true?

dave-ok: While looking for the nearest scrollable element. On these sites mentioned the html and body's height is set to a 100%. When I checked the styling on these sites I noticed that even though the body has a style of `overflow-y: scroll`, the the `html` element gets the scrollable overflow instead. But the scrollable element gotten from the function is the `body` element which does not scroll. I initially made the test to check for the body element but I felt it was not generic enough. In these scenarios the 'body` and `html` scrollable height are also equal. So I leaned towards using that instead

dave-ok:
@ianb On further investigation, I found out that depending on browser the last ancestor scrollable element is either the `<body>` or the `<html>` tag depending on the browser. In the case of firefox it is the `<html>` element but the `<body>` element is not scrollable, neither does it have scrolling properties e.g `scrollTop`. So while searching recursively for a scrollable element in our function, if we get up to the `<body>` element, the test passes because it has a `scrollableHeight` which is greater than the `clientHeight`. But it isn't scrollable in FF. So the last available ancestor is the `<html>` element or the `document.documentElement` which is the main document's scrollable element (document.scrollingElement). So even if the `overflow: scroll` style is applied to the `<body>` element, FF silently passes the style to the `<html>` element. I guess this explains the weird style delegation I tried to figure out earlier. [Source](https://makandracards.com/makandra/55801-does-html-or-body-scroll-the-page)

I have made some changes. Could you please review and let me know if this is in the right direction? Thanks

extension/intents/scroll/scrollHelper.js
ianb: OK, I've noticed it before but hadn't understood what quite happened. I think this body/html fix is more specific and probably more correct as a result.

dave-ok: I am glad this worked. Thank you very much

PR #1452, opened 2020-04-01
remove whitespace around feedback border when starting timer

### LINKED ISSUE #1437: Feedback border on timer
When the timer is displayed there should not be a white boarder around the feedback section.

timer-with-border

Confirm timer

noi5e:
Fixes #1437 

First draft PR...

- The original issue seems to stem from the fact that the extension's `<body>` inherits a margin of 8px. If I set `<body>`'s margin to 0px, it takes care of the whitespace.
- However, this throws off the styling for the footer that says "Say 'cancel,' 'pause,' or 'reset' timer" ?" -- the footer overflows the boundaries of the extension, and creates horizontal & vertical scrollbars
    - Hence the second change, removing margin of -8px for `#timer-footer`.
- Potentially this throws off styling for other scenarios. I tried a few and didn't find anything weird, happy to fix as they come up.

Before changing style for `#timer-footer`:
pull_request

After changing style for `#timer-footer`:
pull_request2


noi5e:
After sleeping on it, I realized a better way to do it would just be to put margin: -8px on #intent-feedback. 

That way I don't have to throw off styling on `<body>`. Duh! So please disregard the previous comment.

Issue #1359, opened 2020-03-21
Add UI to view voice history


ianb:
With #1303 landed we have history being saved, but no way to view it.

For this, add a new tab to the options page, called "History". This should display history items most-recent-first. I believe a table with a formatted date and the utterance would be sufficient.

It would be okay to land this without pagination, but we'll definitely need pagination eventually. Ideally no history will be loaded unless you load the history tab.

@michael-mml would like to do this as a followup to the database work (comment here and I can assign you).

@awallin might have design notes.

michael-mml:
Hi, @ianb , I'd like to take a shot at this.

awallin:
@ianb we should probably allow the user to disable retention of voice history and clear. Are those features I should include in the design?

<img width="703" alt="Screen Shot 2020-03-24 at 9 46 18 AM" src="https://user-images.githubusercontent.com/428245/77453376-6be60900-6db4-11ea-920c-0f69893c547b.png">

ianb:
@awallin Sure, though those could be part of the first PR or not.

We should probably have a note that this is just the local copy of the history. We don't have a way to delete the Telemetry history currently (the Telemetry team is still working on it I think), and if you've disabled Telemetry utterance saving then we might want to include something that says it is saved locally but not uploaded. The combinations are a bit complicated!

michael-mml:
Hi @ianb and @awallin, I'm nearing completion for this feature and realize I will need an icon for the History tab. I don't believe one currently exists.

Also, I had to make a couple of changes to @OliviaRuyinZhang's PR, but we can discuss in the actual PR itself.

Thanks!

awallin:
@michael-mml let's use the standard history icon for Firefox which I've attached.

[History.svg.zip](https://github.com/mozilla/firefox-voice/files/4429524/History.svg.zip)

Issue #1237, opened 2020-03-11
Save history


ianb:
We should save the history of everything you say to Firefox Voice.

This should happen in a new module in `extension/history.js`. `intentRunner.js` would add history items as part of the `runIntent()` function.

The history should be saved using [IndexedDB](https://developer.mozilla.org/en-US/docs/Web/API/IndexedDB_API). In other places we use `browser.storage`, but that's not very scalable for an ever-increasing list.

IndexedDB is not awesome to use, but I'm not excited about adding a library. Still, [Dexie](https://dexie.org/) looks OK, and doesn't have any dependencies from what I see.

To add a new dependency, try to install it with `npm` and then edit `build:deps` in `package.json`. The file (hopefully it's just one file) can be copied into `extension/js/vendor/` – it appears Dexie [supports ECMA modules](https://dexie.org/docs/Tutorial/Consuming-dexie-as-a-module).

After history is saved the next step will be displaying the history to the user.

OliviaRuyinZhang:
Hi @ianb, I would like to give it a try. Could I get this one? Thanks

saucekode:
Hello @ianb can I please work on this. Thank you

ianb:
I'll assign to @OliviaRuyinZhang – we'll be adding more new issues today too, so keep an eye out

OliviaRuyinZhang:
Hi @ianb, I have finished developing a framework for indexedDB. Now, we can create table(s) to store any history we would like to store on browser. I modified intentRunner.js as well to save everything users say to Firefox Voice using indexedDB. 

I did not develope a UI to display the history to users yet since you said that would be the next step. Furthermore, I would like to see if there's any feedback or suggestion about the framework or the requirenments of the UI design before I move on.

YufeiCui:
https://github.com/mozilla/firefox-voice/pull/1303

PR #1453, opened 2020-04-02
Adds history UI component

michael-mml:

Addresses: [#1359](https://github.com/mozilla/firefox-voice/issues/1359)

Also had to make a couple of changes to #1237.


extension/options/history/historyView.js
```diff
@@ -0,0 +1,110 @@
+import { Database } from "./../../history.js";
```
ianb, line 1: Whoops, you have the generated file in here. You should add `historyView.js` to `.gitignore` and delete it from the commit.

michael-mml: Will do!

michael-mml:
Hi Ian, one of our teammates also added a couple of unit tests for the IndexedDB portion which required a mocked IndexedDB library

extension/history.js
```diff
@@ -59,9 +60,9 @@ export class Database {
    * primaryKey), "next" is most recent last (based on primaryKey)
    * @returns {Promise} The documents
    */
-  getAll(tbName, direction) {
+  static getAll(dbName, tbName, direction) {
```
ianb, line 63: It looks like this isn't supported by our tooling. I guess Firefox supports it now? It didn't when I started this but times change quickly!

But, since this makes the linter angry, either that would have to be resolved or this fixed. I'd recommend just rewriting these to `Database.get = function(dbName, tbName, primaryKey) {...};`

extension/history.js
```diff
@@ -1,10 +1,11 @@
 /* globals log */
 
 export class Database {
+  static readonly = "readonly";
+  static readwrite = "readwrite";
```
ianb, line 5: And this probably has to be changed to `Database.prototype.readonly = "readonly";`

michael-mml:
Not sure why CI is complaining about `npm test` since I don't see that on my local machine when I run it.

PR #1455, opened 2020-04-02
example intent for viewPageSource was written wrongly

Simpcyclassy:


ianb:
I _think_ I got this all fixed in #1486 but re-merging won't hurt

Issue #389, opened 2019-10-08
[browser command] Erase history


awallin:
Enable "Erase history", "clear history", or "delete history"

maitrella:
Are you thinking about browser.history.deleteAll() ? In which intents file should be added?

awallin:
@ianb can help with the function needed to call.

Regarding functionality it should bring up the clear history dialog the same as if you clicked on "History > Clear Recent History..." from the browser menu.

<img width="486" alt="Screen Shot 2019-11-25 at 11 43 45 AM" src="https://user-images.githubusercontent.com/428245/69580602-f9395e00-0f78-11ea-837a-6454cdedab6e.png">

maitrella:
I'll wait for @ianb answer but browser.history.deleteAll() erases history but it doesn't bring up the dialog. :-(

ianb:
It seems extreme to immediately erase all history given just that command. But we don't really have patterns for confirmation. I also don't think `browser.history.deleteAll()` quite matches this dialog – i.e., it doesn't delete logins, cookies, cache, or forms.

I don't have a great sense on how best to move forward in this case.

maitrella:
A less extrem option could be browser.history.deleteRange() but it has the same functionality.

awallin:
Agree, we don't want to be destructive without confirmation which is why opening the dialog feels like the right action to take. Is that not possible?

maitrella:
@ianb maybe this API https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/browsingData can help us to find new ways to think about the functionality proposed by @awallin  

What do you think?.

Simpcyclassy:
I just tested it and `Tools:Sanitize` is the command that brings up that dialogue box. Please lemme fix this quickly and go back to my other debugging issues

alexandra-martin:
Verified on Mac 10.14 with Firefox Nightly 77.0a1 (2020-04-08).

![1](https://user-images.githubusercontent.com/54401091/78784449-072ecf00-79ae-11ea-919b-55136554ce5a.gif)

PR #1456, opened 2020-04-02
[browser command] Erase history

Simpcyclassy:
Please I need your reviews @ianb @awallin

### DIFF
--- a/extension/experiments/voice/api.js
+++ b/extension/experiments/voice/api.js
@@ -55,6 +55,10 @@ this.voice = class extends ExtensionAPI {
             return runCommand("Tools:Downloads");
           },
 
+          async clearBrowserHistory() {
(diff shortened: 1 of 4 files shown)


ianb:
@awallin do you have thoughts on this? It brings up this dialog:

image

Simpcyclassy:
From the issue explanattion here https://github.com/mozilla/firefox-voice/issues/389#issuecomment-558353897 I believe it is the expected behaviour

awallin:
Yes, that's the expected behavior. It'll be nice when we can confirm the action with "Ok" as a voice command.

ianb:
OK. Hitting "OK" will be very hard, but eh. If we implemented our own history clearing (there are extension APIs) we could do something like "clear today's history", with our own confirmation.

PR #1460, opened 2020-04-02
Fixes #1424 - Album support deezer

### LINKED ISSUE #1424: Add album/playlist search to Deezer
#1293 added album and playlist support to Spotify and YouTube, it could also be added to Deezer

melvin2016:
**Outreachy applicants:** please try to get all _new_ PRs submitted by Friday,
April 3rd. After that we'll be trying to review all PRs and respond to 
followups. Earlier is better, as the review load is very high.

Fixes #1424 .


melvin2016:
@ianb Please review :)

extension/services/deezer/player.js
```diff
@@ -1,52 +1,68 @@
...
+          }
+        } catch (e) {
+          throw new Error("No Search Results!");
         }
       } catch (e) {
-        const unlogged = this.querySelector("div[class='unlogged-homepage']");
-        if (unlogged) {
-          throw new Error("Please log in to use this service.");
-        }
+        throw new Error(e.message);
```
ianb, line 39: This doesn't do anything any longer (except break the traceback by recreating the exception). But it also looks like this code isn't right. It should probably be:

```js
...} catch (e) {
  const unlogged = document.querySelector(".unlogged-homepage");
  if (unlogged) {
    throw new Error("Please log in to use this service.");
  }
  throw e;
}
```

extension/services/deezer/player.js
```diff
@@ -124,6 +140,39 @@ this.player = (function() {
         iconVolumeParent.click();
       }
     }
+
+    async playSection({ query, thenPlay, section }) {
+      try {
+        let SECTION;
```
ianb, line 146: This shouldn't be an upper class variable, that should only be used for constants or global config variables, not a mutable local variable. Maybe `foundSection` would be a good name?

extension/services/deezer/player.js
```diff
@@ -124,6 +140,39 @@ this.player = (function() {
...
+        await this.search(query);
+        try {
+          if (thenPlay) {
+            const anchorNodes = await this.waitForSelector(".container h2 a", {
+              all: true,
+              timeout: 5000,
+            });
+            for (const anchorTag of anchorNodes) {
+              if (anchorTag.innerText === section) {
+                SECTION = anchorTag.parentElement.parentElement;
```
ianb, line 156: This should probably `break;` once it is found

extension/services/deezer/player.js
```diff
@@ -124,6 +140,39 @@ this.player = (function() {
...
+              }
+            }
+            SECTION.querySelector("ul li button").click();
```
ianb, line 159: This needs to check if `SECTION` is still undefined. Probably it should just raise an exception saying that the section couldn't be found.

extension/services/deezer/player.js
melvin2016: Oh. Sorry. Yes it was taking much time searching for all the sections thus making it slow.

extension/services/deezer/player.js
melvin2016: Ok

extension/services/deezer/player.js
ianb: If it always re-throws the error, then the try/catch isn't necessary at all.

melvin2016:
> Looks good, thanks! There's one issue with try/catch that I'll just fix in the interest of getting this landed.

Thank You! :)

PR #1461, opened 2020-04-02
fix bug displaying right result on doorhanger

### LINKED ISSUE #1397: Doorhanger with wrong result and a new tab with right result open at the same time when making a command
### Prerequisites:
Mic permissions are enabled.
"A privacy reminder from Google" is reviewed and agreed and the yellow language pop-up is closed. (if applicable)

### STR:
2. Say or write commands that should open cards in the doorhanger, e.g. "lion".
3. Press the doorhanger "Back" button or close the doorhanger and open it again using the shortcut or mic icon from the browser toolbar.
4. Say or write the "56 diamonds" command.

### Expected result:
Result is either displayed in the doorhanger or a new tab opens with the result (Google or website).

### Actual result:
A new tab opens with a website while the doorhanger is displaying the result from the previous command with the title of the current command.

### Notes:
Reproduced on Mac 10.14 with Firefox Nightly 76.0a1 (2020-03-27).
The issue was reproducing with other similar commands, but it was constantly changing. So the commands mentioned in the steps might not reproduce the issue after a while.

danielamormocea:
Fixes #1397

@ianb I hope this fixes the bug reported in #1397. I think the problem was that the cardImage was not cleared sometimes.

### DIFF
--- a/extension/popup/popupController.jsx
+++ b/extension/popup/popupController.jsx
@@ -386,6 +386,8 @@ export const PopupController = function() {
     if (message.card) {
       setCardImage(message.card);
       setMinPopupSize(message.card.width);
+    } else {
+      setCardImage(null);
     }
 
     updateLastIntent();


ianb:
This change makes sense to me as a fix. Thanks!

PR #1462, opened 2020-04-02
added search and say intents for linguee

annlinros:
@ianb please review this PR for #1299 

search [query] on linguee works,

say [query] on linguee almost works, except that the click event is not making the audio to play, instaed a dialog box opens to select british/american english. 

And in the browser console I get this error - 
> The play method is not allowed by the user or the platform in the urrent context, possibly because the user denied permission

Screenshot 2020-04-02 at 19 39 17

### DIFF
--- a/extension/intents/linguee/linguee.js
+++ b/extension/intents/linguee/linguee.js
@@ -1,0 +1,20 @@
+/* globals */
+
+import * as intentRunner from "../../background/intentRunner.js";
+import * as searching from "../../searching.js";
+import * as content from "../../background/content.js";
+
+intentRunner.registerIntent({
+  name: "linguee.say",
+  async run(context) {
+    const myurl = await searching.ddgBangSearchUrl(
+      context.slots.query,
+      "linguee"
+    );
+    const lingueeTab = await context.createTab({ url: myurl });
+    await content.lazyInject(lingueeTab.id, [
+      "/intents/linguee/lingueeContentScript.js",
+    ]);
+    await browser.tabs.sendMessage(lingueeTab.id, { type: "sayWord" });
+  },
+});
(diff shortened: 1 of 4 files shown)


annlinros:
@ianb 
I have then tried doing button.onclick() and dispatched a click MouseEvent..

I also tried 
` const clickEvent = new MouseEvent("click", {
      bubbles: true,
      view: window,
      cancelable: true,
    });

const loudspeaker = document.querySelector(".loudspeaker");
const parent = loudspeaker.parentElement;
parent.dispatchEvent(clickEvent);`

but nothing seems to work.
Although when I do button.click() on the browser console, it plays the audio.

Why does the triggering for a click event not work?
Could you please give me any tips? Thanks.

dave-ok:
@annlinros I tested your code and it seems fine. After some digging I found that the issue is a firefox policy that prevents automatically playing audio on a webpage, until there is a user interaction involving at least a click or keypress. Supposedly, it prevents sites from just playing annoying audio as soon as they load unless given such permissions. Just running the intent on a freshly loaded page won't work also. I tried to fake a keypress event first on some temporary inputElement but no joy. Maybe you might find a better way to fake user interaction before the clicking the speaker button. 
I found some useful resources here if you want to give it another go:
https://stackoverflow.com/questions/57504122/browser-denying-javascript-play
https://searchfox.org/mozilla-central/rev/a650fac7b069ded3d63eba530271ce80e6cafb1b/widget/TextEvents.h#303
https://developer.mozilla.org/en-US/docs/Web/API/HTMLMediaElement/play

annlinros:
@dave-ok  I see! Thank you so much..I was so confused as to why it doesn't play the audio when I was doing everything right. Now I understand why my "play music" intents also doesn't get played unless I click on the play button or say play again..I was thinking that was the desired behaviour until now..
I will check the links. Thanks again Dave!

PR #1464, opened 2020-04-02
Timer controls 'Cancel', 'pause', and 'reset'

Simpcyclassy:


extension/intents/timer/timer.toml
```diff
@@ -11,40 +11,55 @@ phrase = "Set timer for 2 hours and 5 minutes"
 [timer.close]
 description = "Close timer"
 match = """
+<<<<<<< HEAD
```
ianb, line 14: Looks like you added some conflict markers

ianb:
@awallin this would change "cancel" to stop a timer, if there is a timer. I'm not sure we want to do that, but I'm not sure; thoughts?

If it does cancel the timer, then I think we'd want it to show a confirmation message (and so it also shouldn't close immediately).

extension/intents/timer/timer.toml
Simpcyclassy: My bad. Will fix that now

awallin:
`Cancel` should stop the timer. And yes, we should confirm that with the checkmark with "Timer Canceled" below.

Simpcyclassy:
@ianb may I get pointers on how to add this confirmation message. I tried a `context.displayText()` and it gave no result.

Secondly for the pause aspect, if I were to use same implementation I used for cancel and there are a couple of matching intents for `music.pause` it would mean that an utterance like `pause music` will pause timer, I do not think that will be ideal.

Issue #1135, opened 2020-03-02
[wakeword] investigate downsampler errors


ianb:
I'm getting errors like:

```
TypeError: can't access property "postMessage", downsampler is undefined 2 web_voice_processor.js:39:9
```

This appears to break the wakeword handling.

Simpcyclassy:
@ianb Can I investigate this please?

ianb:
@Simpcyclassy It looks like you have several issues now, so leaving this for now.

Simpcyclassy:
OK, that's fine.

ianb:
No longer applicable since we removed the underlying library.

PR #1471, opened 2020-04-03
Wakeword

danielamormocea:
Draft for #1135

I looked into this issue and tried to figure it out. I know that the final result might not be satisfactory, but I've been through several fixes and none seemed to work well so far. 

I realised that the problem was that everytime the error you mentioned appeared, the wakeword stopped working (the user might have to go to the extension tab and refresh it, but even then, sometimes might work, sometimes might not). I placed a try catch around the line that caused the problem to ensure that the code reaches the part that starts the recorder.

I also tried to start the recorder everytime the browser was opened by focusing the wakeword.html tab and refreshing it so that the recorder can load, but then again, this sometimes works and sometimes doesn't (sometimes even after the reload, the user has to interact with the page for the recorder to start). 

Additionally, I tried to place a delay when openWakeword is called in main.js in case wakeword.html wasn't loaded in time. 

@ianb  please take a look at this and tell me if you have a recommendation for another approach

### DIFF
--- a/extension/recorder/wakeword.js
+++ b/extension/recorder/wakeword.js
@@ -63,7 +63,7 @@ function stopWatchword() {
   try {
     porcupineManager.stop();
   } catch (err) {
-    log.error(`Error in porcupine Manager ${err}:`, err);
+    log.error(err);
   }
   log.info("Stopped listening for watchwords");
   enabled = false;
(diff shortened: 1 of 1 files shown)


ianb:
We weren't able to figure out how to proceed with Porcupine to use our own wakeword, so we're removing this functionality (at least the part applicable here) in #1597.

So this doesn't apply anymore... but thanks for putting in some effort here!

Issue #1479, opened 2020-04-05
Fixes #905 - Tab selection 


shreyajain25:
"Select all tabs to the left" 
"Select all tabs to the right"
"Select the last three tabs"
"Select the first two tabs"
"Select all the tabs"
Selecting multiple tabs according to the queryies

#1481 

extension/intents/tabs/tabs.js
extension/intents/tabs/tabs.toml

PR #1481, opened 2020-04-05
Fixes #905 - Tab selection #1479

### LINKED ISSUE #905: Tab selection
Based on telemetry users are interested in more advanced tab management utterances such as:

"Select all tabs to the left" (would close all tabs left of the active tab)
"Select Lifehacker tab" (would only close the matching tab)
"Select the last three tabs"
"Select the first two tabs"
"Select all tabs except Spotify"
"Select all tabs without gmail"
"Select all the tabs"
"Select all tabs but this one"

shreyajain25:
Files changed: 
  extension/intents/tabs/tabs.js
  extension/intents/tabs/tabs.toml

1. Added select all the tabs
2. Added select tabs to the right/left
3. Added select first/last three tabs (any number of tabs)


extension/intents/tabs/tabs.js
```diff
@@ -597,3 +598,166 @@ intentRunner.registerIntent({
...
+    await selectAllTabs(matchingTabs);
+
+  },
+});
+
+async function selectAllTabs(tabs){
+  if(tabs.length !== 0){
+    browser.tabs.query({
+      currentWindow: true
+    }).then((tabs) => {
```
ianb, line 633: This will be easier to read (and safer in case of exceptions) if you use `await`, like:

```js
tabs = await browser.tabs.query({currentWindow: true});
```

Also this should also use `pinned: false`, as most things that handle "all" tabs only handle the not-pinned tabs.

extension/intents/tabs/tabs.js
```diff
@@ -597,3 +598,166 @@ intentRunner.registerIntent({
...
+      for (var tab of tabs) {
+        if (!tab.active){
```
ianb, line 635: There's a lot of formatting and lint errors here. `npm run format` will help with formatting, and `npm run lint:js` will show you lint errors (like using `var` instead of `const` or `let`).

extension/intents/tabs/tabs.js
```diff
@@ -597,3 +598,166 @@ intentRunner.registerIntent({
...
+          await browser.tabs.update(tabId, {
+              highlighted: true,
+              active: false,
```
ianb, line 638: Does this need to specify `active` at all?

extension/intents/tabs/tabs.js
```diff
@@ -597,3 +598,166 @@ intentRunner.registerIntent({
...
+          await browser.tabs.update(tabId, {
+            highlighted: true,
+        });
+        }
+      }
+    });
+  }
+}
+
+async function highlightSelectedTabs(center, tabs, dir) {
```
ianb, line 650: I think `highlightTabsInDirection` or something like that would be a better name.

extension/intents/tabs/tabs.js
```diff
@@ -597,3 +598,166 @@ intentRunner.registerIntent({
...
+  if(dir === "left"){
+    const tabsLeft = tabs.filter(tab => {
+      return tab.index < center.index;
+    });
+    for (let i = 0; i < tabsLeft.length; i++) {
+      tabLeft = tabsLeft[tabsLeft.length - i - 1];
```
ianb, line 656: Is there a reason you are going from the left over? It seems like this could just be `for (const tab of tabsLeft) { await browser.tabs.update(tab.id, {highlighted: true})`

extension/intents/tabs/tabs.js
```diff
@@ -597,3 +598,166 @@ intentRunner.registerIntent({
...
+          highlighted: true,
+        });
+      }
+    }
+  }
+  else if(dir === "right"){
+    const tabsRight = tabs.filter(tab => {
+      return tab.index > center.index;
+    });
+    for (let i = 0; i < tabsRight.length; i++) {
```
ianb, line 673: and it seems like you don't need different logic for the right and left highlighting, you only need it for selecting the right tabs.

extension/intents/tabs/tabs.js
```diff
@@ -597,3 +598,166 @@ intentRunner.registerIntent({
...
+  }
+
+}}
+
+intentRunner.registerIntent({
+  name: "tabs.highlightSelectedTabs",
+  async run(context) {
+    const activeTab = await context.activeTab();
+    const tabs = await browser.tabs.query({ currentWindow: true});
+    const allWindows = await browser.windows.getAll({
```
ianb, line 694: It doesn't look like this is used

extension/intents/tabs/tabs.toml
```diff
@@ -377,7 +377,49 @@ match = """
 
 [[tabs.findOnPage.example]]
 phrase = "Find query on this page"
-test = true
```
ianb, line 380: This should stay as a test only

extension/intents/tabs/tabs.toml
```diff
@@ -377,7 +377,49 @@ match = """
...
+[[tabs.selectAllTabs]]
+description = "Select all tabs"
+match = """
+    (select | highlight) (all |) (the | this ) (tab{s} | taps | page )
+"""
+[[tabs.selectAllTabs.example]]
+phrase = "Select all the tabs"
+
+[[tabs.selectAllTabs.example]]
+phrase = "Select all tabs"
```
ianb, line 393: One of these two examples should be marked as `test = true`

extension/intents/tabs/tabs.toml
```diff
@@ -377,7 +377,49 @@ match = """
...
+
+[[tabs.highlightSelectedTabs]]
+description = "Select tabs to the left/right"
+match = """
+    (select | highlight) all (tab{s} | page) (to | in | by |) (the |) right [dir = right] [allWindows = false]
+    (select | highlight) all (tab{s} | page) (to | in | by |) (the |) left  [dir = left] [allWindows = false]
```
ianb, line 399: I'd use `(all |)` to make that optional, and `page{s}` to match both page and pages.

extension/intents/tabs/tabs.toml
```diff
@@ -377,7 +377,49 @@ match = """
...
+
+[[tabs.highlightSelectedTabs.example]]
+phrase = "Select all tabs to the left"
+expectParameters = {dir = "left"}
+
+[[tabs.selectNumbersTabs]]
+description = "Select first/last numbers of tabs"
+match = """
+    (select | highlight) first [number] (tab{s} | taps | page ) [dir=first] [allWindows = false]
+    (select | highlight) last [number] (tab{s} | taps | page ) [dir=last] [allWindows = false]
```
ianb, line 414: You could use `[number:smallNumber]` to only match numbers.

extension/intents/tabs/tabs.js
shreyajain25: Capture111

Highlight : (Adds or removes the tab from the current selection. If true and the tab is not highlighted, it will become active by default.

If you only want to highlight the tab without activating it, Firefox accepts setting highlighted to true and active to false. Other browsers may activate the tab even in this case.)


Actually, we don't have to activate all the tabs , but only highlight them. So, whatever tab is active will be active other tabs wont be active only highlighted.
So, I think it is needed

ianb:
I've added a few things to this PR in #1743 (closing here to merge that)

Issue #1487, opened 2020-04-07
Add "maximize window" alias


ianb:
#1434 adds "zoom window", but I think "maximize window" should also be allowed.

farhatcode:
May be fixed by #1492

PR #1492, opened 2020-04-07
Adds maximize window

### LINKED ISSUE #1488: Add maximize support to Windows
I'm pretty sure the "zoom window" command in #1434 is Mac-specific (at least my search in the code seemed to show that). We should find an equivalent way to do it on WIndows.

farhatcode:
Fixes #1488


farhatcode:
May also resolve #1487

ianb:
Ah, cool, I missed that the API existed for this!

Issue #1363, opened 2020-03-23
"Turn selection into link" gives "No text selected" when used in Google Docs and Google Slides


alexandra-martin:
### Prerequisites:
Mic permissions are enabled.
"Allow" or "Don't Allow" button from “Allow Firefox Voice to Collect Voice Transcriptions” pop-up is clicked.
A paragraph is copied in Google Docs, or Google Slides and the tab is made active. 

### STR:
1. Select a string from the copied paragraph. (it can be a word or multiple words)
2. Use the shortcut or click on the mic icon from the browser toolbar.
3. Say or write the "Turn selection into link" command.

### Expected result:
The selected string is turned into a link and can be accessed if clicked on it.

### Actual result:
"No text selected" is displayed in the doorhanger.

### Notes:
Reproduced on Mac 10.14 and Win10x64 with Firefox Nightly 76.0a1 (2020-03-23).
Not reproduced, if text is copied in Google Sheets or [Calmly Writer](https://www.calmlywriter.com/online/). 

![2](https://user-images.githubusercontent.com/54401091/77327755-52957c00-6d24-11ea-844d-c1685b09d5a6.gif)

noi5e:
I'd like to work on this if possible. I've started to do research in the codebase, would appreciate any leads as well!

PR #1493, opened 2020-04-07
Google Docs: Retrieve Selected Text

noi5e:
This is a draft PR to fix Issue #1363. 

This PR adds a little functionality to the extension, but it really is only halfway there. I'm just making a draft PR now to get a little guidance and direction (and yes, to get some extra credit for the Outreachy app, thought it couldn't hurt 🤷‍♂).

- The issue stems from Google Docs being complicated. You can't simply run document.getSelection() to retrieve text because Google Docs wraps every line & word in its own DOM node.
- Luckily [this Chrome extension](https://github.com/JensPLarsen/ChromeExtension-GoogleDocsUtil/blob/master/googleDocsUtil.js) provides some pointers as to how to work with Google Docs selections. I refactored relevant parts of this code as helper functions in pageMetadata-contentScript.js.
- In this PR, Voice is able to detect selected text in Google docs (logged to the console in the .gif below). Voice automatically opens up a Google search tab with the selected text. If bing.com were to be transformed into a link, Voice would direct you to Google, which would then redirect you to bing.com (as if you entered "bing.com" into a Google search).

pull_request

If I were to follow through with this approach, the next step would be to actually transform the selected text into a usable link within the Google Docs page. Not sure how to do this yet. Maybe the easiest way is to simulate the relevant keypresses or clicks?

Before I go any further, just had some questions:
- the helper functions in pageMetadata-contentScript.js are huge, and probably need to be split into another file. what structure should i follow?
- also made slight changes to pageMetadata.js & forms.js, please take a look.
- general feedback on if this is a good direction to head in, or other possible approaches.


ianb:
@noi5e My goodness, that is some large and very complicated code!

It could get put in another script and lazyloaded. I'd suggest loading it in `extension/background/pageMetadata.js` but unfortunately sometimes we load the content script directly without using that function. So one option would be to extend the `getSelection()` function to read the URL and add this script if necessary, and to review any cases where we are using `lazyInject([".../pageMetadata-contentScript.js"])`

Is it possible for a script to call `browser.tabs.executeScript` from a content script? Probably not, but that would be another approach.

Or, update `getSelection()` to inject the script when appropriate, and just let it fail gracefully if it's not present. That's probably the easiest approach? We inject `pageMetadata-contentScript` frequently so I don't want it to get too large.

PR #1495, opened 2020-04-07
Close All Tabs Resolves #1480

### LINKED ISSUE #1480: Search "Close All Tabs" commands make a new Search
Open The fresh Browser 
 Speak "Close All Tabs"

Actual Response : Makes a new Search

awuorm:
**Outreachy applicants:** please try to get all _new_ PRs submitted by Friday,
April 3rd. After that we'll be trying to review all PRs and respond to 
followups. Earlier is better, as the review load is very high.

### DIFF
--- a/extension/intents/tabs/tabs.toml
+++ b/extension/intents/tabs/tabs.toml
@@ -12,9 +12,6 @@ phrase = "close tab"
 phrase = "close tabs"
 test = true
 
-[[tabs.close.example]]
-phrase = "close all tabs"
-test = true
 
 [[tabs.close.example]]
 phrase = "close active tab"
--- a/extension/intents/window/window.toml
+++ b/extension/intents/window/window.toml
@@ -17,6 +17,9 @@ phrase = "open downloads"
 description = "Close Firefox"
 match = """
   (close | quit | exit | shutdown) (firefox | browser | the browser)
+  close all tabs
+  close all tab
+  all tabs close
 """
 
 [[window.quitApplication.example]]


awuorm:
The voice command "close all tabs" now quits firefox while initially it would just close the current tab only.

ianb:
@awallin What do you think we want to do in this case?

awallin:
Not sure we want to quit in this case. Quitting Firefox seems like it could have the possibility of a destructive consequences and the user may want to follow up with another action.

Is it not possible to close all the tabs without quitting?

ianb:
@awallin we can close all tabs, but it's kind of inconsistent what happens then. If there's multiple windows then it's like closing the window. If there aren't, it will exit Firefox on Windows, and on Mac leave it in a weird no-window-still-open state. And then there's the question of whether pinned tabs would be included.

Again, it's all because "all" is a stopword and the phrase isn't rejected outright, but is executed in a way that doesn't match the phrase. Maybe we should have explicit non-matching phrases?

awuorm:
@ianb when I was working on this, I noticed that the "close all tabs" phrase was under close tabs, hence why only the active tab was being closed. From my understanding, when a user wants to close all tabs it is equivalent to exiting the browser window. Would the other option of cascading through the active tabs till all the tabs are closed be more desirable in this case?

ianb:
I think the underlying ticket has been closed, and we don't really know what the preferred behavior is, so there's no right solution here. I'm going to close this for now, please reopen if you think there's something we should continue with here.

PR #1514, opened 2020-04-09
Search in hidden tab issue

### LINKED ISSUE #1475: Search page is reused for fallback searches
I think this regressed with some of the recent search changes, but there's been too many for me to be certain where or when.

To reproduce:

1. Say "search for weather"
2. Say "weather in boston"

This *should* open a search page for 1, and then do a search in a hidden page for 2, and use the result in 2 to show a card. At this moment the card doesn't display, but that is tracked in #1474.

What *does* happen is that the search page opened in 1 is also used for the search in 2. When the card is displayed it's therefore superfluous, since the main search page also is displayed.

I'm assuming this is a problem with how `_searchTabId` is managed in `intents/search/search.js`

danielamormocea:
Fixes #1475

This PR reuses the tab only if it's hidden to avoid both showing the card and searching in page (when the card is available).

### DIFF
--- a/extension/intents/search/search.js
+++ b/extension/intents/search/search.js
@@ -36,8 +36,11 @@ async function openSearchTab() {
   }, CLOSE_TIME);
   if (_searchTabId) {
     try {
-      await browser.tabs.get(_searchTabId);
-      return _searchTabId;
+      const tab = await browser.tabs.get(_searchTabId);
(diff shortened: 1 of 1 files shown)


ianb:
I suspect there's another logic error in how `_searchTabId` is handled (maybe it is set sometimes when it shouldn't be), but this will at least help. Thanks!

PR #1515, opened 2020-04-09
phrase and example fixes

farhatcode:
After syncing with latest master, some phrase tests were crashing. 
This PR contains some simple phrase and example fixes to resolve that.

1. One of the phrase match in `[tabs.undoCloseTab]` was colliding with `[tabs.open]`. Therefore removing the keyword 'open' from `[tabs.undoCloseTab]` intent match and also removing the relevant example. 
2. Removing an example from `[browser.preferences]` intent that didn't fulfil the intent match and therefore caught by `[self.openOptions]`.
3. Fixing a misspell in an example in `[window.clearBrowserHistory]` intent, due to which example was unable to match with any intent.

### DIFF
--- a/extension/intents/browser/browser.toml
+++ b/extension/intents/browser/browser.toml
@@ -67,9 +67,6 @@ match = """
 """
 
 [[browser.preferences.example]]
-phrase = "open settings"
-
-[[browser.preferences.example]]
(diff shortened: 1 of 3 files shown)


ianb:
Thank you!

Issue #1469, opened 2020-04-03
"Increase\Decrease volume" not supported on Spotify


alexandra-martin:
### Prerequisites:
Mic permissions and autoplay are enabled.
"Allow" or "Don't Allow" button from “Allow Firefox Voice to Collect Voice Transcriptions” pop-up is clicked.
A Spotify account is logged in, the Spotify tab is active and music is playing.

### STR:
1. Use the shortcut or click on the mic icon from the browser toolbar.
2. Say or write the "Increase volume" or "Decrease volume" command.

### Expected result:
Volume is adjusted accordingly.

### Actual result:
"Internal error: Error: No handler for adjustVolume" is displayed in the doorhanger.

### Notes:
Reproduced on Mac 10.14 with Firefox Nightly 76.0a1 (2020-04-03).

![6](https://user-images.githubusercontent.com/54401091/78376075-75eae180-75d6-11ea-82d9-4873dfbb439e.gif)

farhatcode:
@alexandra-martin, I have added the volume controls for youtube, soundcloud and deezer in PR #1375 .
Volume controls for spotify are not added yet. You can see this here:
https://github.com/mozilla/firefox-voice/issues/1095#issuecomment-606289061

annlinros:
@alexandra-martin I am working on Spotify's volume control..

alexandra-martin:
@farhatcode thank you for the work done so far. I saw your comment that you don't have Spotify available, which is why I wrote the report so someone else can help with that. I should have mentioned the previous report in notes, sorry about that.
The new or follow-up reports logged are not meant to be made in a critical way, as we appreciate every help we get and by collaborating we ensure that we have a great product at the end.

PR #1525, opened 2020-04-12
1469 volume controls spotify

annlinros:
Draft PR for #1469 

Hi @ianb 
I have implemented mute, unmute actions for Spotify.
But with the increase volume intent, 
like Deezer or SoundCloud, there was no DOM element with an aria label-now attribute to get the current volume from. And like Youtube, there was also no <audio> element to manipulate volume change.
I had tried dispatching a keypress event on the slider and also a mouse-down event on the progress-bar, none of which seems to trigger a volume change.

I really do need your help on this. Thank you.

### DIFF
--- a/extension/services/spotify/player.js
+++ b/extension/services/spotify/player.js
@@ -1,4 +1,3 @@
-/* eslint-disable no-undef */
 /* globals helpers */
 
 this.player = (function() {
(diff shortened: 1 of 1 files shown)


ianb:
Looks like this is out of date, please reopen if you want to continue with it

Issue #1166, opened 2020-03-06
Home page for Firefox Voice


awallin:
To date Firefox Voice has been using Open Innovation's campaign tools. As we expand the audience we'll need a more established home page for promotion that can highlight functionality, provide answers to common questions (which we get a lot of), and a link to download the add-on.

This will also be a location to display #643

saucekode:
@awallin when this is ready. Please, I would love to work on it. Thank you.

awallin:
Design for a new home page is ready. This new home page is focused on two audiences.

1. Existing [Common Voice](http://voice.mozilla.com) contributors. Our hope is they'll be excited to use Firefox Voice and be open to allowing their voice recordings to be saved and used as training data for Deep Speech’s SST model.

2. General public who discovers the extension through the add-on store, promotion in Firefox or other press/marketing.

Review the UX diagram for details on how the UI/flow change depending on referral source.

Details and specs for the design can be found here:
* Attached JPG mockups at various responsive breakpoints
* [HTML and CSS specs can be found on Abstract](https://share.goabstract.com/23673cdc-f2e3-405d-ae7a-70e0a286f278)
* [Content is available in this Google Doc](https://docs.google.com/document/d/17MnBTuO4Mxl7HZbYF6nxMVtni-MzuYLHkjtauvKiqxs/edit?usp=sharing)

From a technical perspective this site should be implemented using Mozilla's Protocal design system(https://protocol.mozilla.org/). @ianb can you add details about how/where the site/repo should be hosted?

![image](https://user-images.githubusercontent.com/428245/78063651-6acd5280-7345-11ea-85e3-46ef9c96cf48.png)

[FxVoice homepage.zip](https://github.com/mozilla/firefox-voice/files/4411010/FxVoice.homepage.zip)

Simpcyclassy:
I believe these issues are same https://github.com/mozilla/firefox-voice/issues/126

saucekode:
Yea sure. Still interested!

jcambre:
@saucekode are you still working on this? I think we're hoping to have the new home page implemented pretty soon, so I'm happy to take it on if you don't have the time in the next few days.

saucekode:
Hey dear, sure. I'm currently on a project. Can we kick off by say Monday?

awallin:
Here's an InVision prototype for the homepage.
https://mozilla.invisionapp.com/share/9XWU20C2DVB#/413305292_Home_-_1056

[...]

PR #1541, opened 2020-04-14
Draft implementation of the homepage (Addresses #1166) 

jcambre:
First page of the homepage / installation flow

- [ ] Reference the issue you are fixing in at least one of your commit messages, as `Fixes #X`


awallin:
This is looking amazing. Thanks @jcambre. A few fixes/clean up, many of which you probably already know about.

* For top navigation items (How it works, FAQs) can we smooth scroll to their location?
* Top nav should become a hamburger menu at smaller sizes
* Link to "Everything you can say" should go to [Lexicon](https://mozilla.github.io/firefox-voice/lexicon.html)
* Clicking on video should open a large size version of the video in a lightbox and begin auto playing.
* I noticed and fixed a couple typos for the first FAQ `What is Firefox Voice?`. Can you copy and paste the answer again?
* Adding the correct links for the footer (in case you don't have them)
** [Installation instructions](https://docs.google.com/document/d/1nTyIw4G1yWnxlTmJ-Nvs9OIDntAIYmRPMLKVJZ9TxkI/edit)
** [Github](https://github.com/mozilla/firefox-voice/)
** [Feedback](https://firefox-voice-feedback.herokuapp.com/)
** [Privacy](https://mozilla.github.io/firefox-voice/privacy-policy.html)

.eslintignore
```diff
@@ -7,6 +7,8 @@ extension/options/routinesView.js
 extension/options/history/historyView.js
 extension/onboarding/onboardingController.js
 extension/onboarding/onboardingView.js
+extension/homepage/homepageController.js
+extension/homepage/homepageView.js
```
ianb, line 11: I think this should all go in `homepage/` (besides, not in, `extension/`). And then we can also probably build it separately. Potentially we should even create `homepage/package.json` and `homepage/public/index.html` etc. Then the build steps can be different. But I'm also fine having the root `package.json` build it.

extension/homepage/homepage.scss
```diff
@@ -0,0 +1,226 @@
+@use '../css/vendor/protocol/protocol';
```
ianb, line 1: This is where the error is happening. Maybe you put those files into place locally but it's not part of the build? Or the build might be happening in a different order, so those files haven't been copied in. I suppose this might also be a way to use that sass without copying it at all: https://www.npmjs.com/package/node-sass-package-importer

awallin:
Attaching image assets.

[FxVoice-homepage.zip](https://github.com/mozilla/firefox-voice/files/4489777/FxVoice-homepage.zip)

ianb:
@jcambre I've moved the files around so the homepage is its own package and top-level directory. It gets published to https://mozilla.github.io/firefox-voice/homepage/ (I put the current version there as well).

PR #1551, opened 2020-04-16
display message instead of internal error for saveAsPDF

farhatcode:
`tabs.saveAsPDF()` is not supported for mac OS and it shows internal error in doorhanger.
This PR adds try catch block to display message to the user instead of displaying internal error.


ianb:
Thanks!

PR #1553, opened 2020-04-16
[Alias] "View bookmarks sidebar" now dispalys bookmark sidebar

Simpcyclassy:


ianb:
@awallin do you have thoughts on allowing bare-word phrases like "history" and "bookmarks" open the respective sidebars?

awallin:
@ianb Is your question does it make sense for the single word "history" or "bookmarks" to open the sidebar as the meaning is not clear (e.g. maybe the user wants to search for those)?

If that's the question I think the most likely scenario is a user is trying to open those items in the browser and so it makes sense to open the sidebar. Although, for `bookmarks` we should be able to distinguish between `bookmark` which is probably an intent to bookmark the page.

ianb:
@awallin I'd be concerned about making "bookmark" and "bookmarks" do different things – speech recognition accuracy is really low for pluralization, especially of a single word, so I think these should be treated the same. But that doesn't clarify what they should do ;)

One possibility would be to make `bookmark{s}` open the bookmark sidebar, but also display "say 'bookmark this' to bookmark the current page"

awallin:
Yea, that makes sense. So go with:

- `Bookmark` and `bookmarks` will open the sidebar.
- `Bookmark this`, `bookmark this page`, `bookmark this site`, etc will add a bookmark for the active page.

Simpcyclassy:
> Yea, that makes sense. So go with:
> 
> * `Bookmark` and `bookmarks` will open the sidebar.
> * `Bookmark this`, `bookmark this page`, `bookmark this site`, etc will add a bookmark for the active page.

I have fixed these accordingly with my new commits

ianb:
Thanks!

PR #1557, opened 2020-04-16
fixes #1548 "Zoom window" gives "Internal error

### LINKED ISSUE #1548: "Zoom window" gives "Internal error: Error: An unexpected error occurred" when made on Windows
### Prerequisites:
Mic permissions are enabled.

### STR:
2. Say or write the "Zoom window" command.

### Expected result:
If the command is not supported on the used device, a message stating that is displayed in the doorhanger together with the correct command that needs to be used on that device.

### Actual result:
"Internal error: Error: An unexpected error occurred" is displayed in the doorhanger.

### Notes:
Reproduced on Windows 10x64 with Firefox Nightly 77.0a1 (2020-04-15).
Not reproduced on Windows for the "Maximize window" command.
on Mac 10.14 both "Zoom window" and "Maximize window" are working as intended.

Simpcyclassy:
It appears the browser runCommand "zoomWindow" does not work on windows. I do not have this tested on a windows machine but from the issue raised in #1548, "maximize window" calls the command which changes windows state to maximized. This works on a mac, I believe an alias which calls the same command "maximize window" calls fixes it for a windows machine

### DIFF
--- a/extension/experiments/voice/api.js
+++ b/extension/experiments/voice/api.js
@@ -86,10 +86,6 @@ this.voice = class extends ExtensionAPI {
             return runCommand("View:PageSource");
           },
 
-          async zoomWindow() {
-            return runCommand("zoomWindow");
-          },
-
           async minimizeWindow() {
             return runCommand("minimizeWindow");
           },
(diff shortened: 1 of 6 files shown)


ianb:
Thanks, this makes sense – I don't know what the subtle differences between "zoom window" and "maximize window" are, but Mac window management has always left me confused. I agree just using the browser's maximize function makes sense.

PR #1378, opened 2020-03-25
Corrects spotify play for new tab

GangaChatrvedi:
Fixes   #1339 

Before submitting a final PR, please:

- [x] Run `npm test` on your machine
- [x] Run `npm run format` to keep code formatted consistently
- [x] Reference the issue you are fixing in at least one of your commit messages, as `Fixes #X`

PR #1564, opened 2020-04-18
Fix regression in #1378

### LINKED ISSUE #1365: "Open Google" blocks other actions if placed first in a routine

### LINKED ISSUE #1526: Commands that need to go through "Redirect Notice" to open a page are not working properly
### Prerequisites:
Mic permissions are enabled.

### STR:
2. Say or write the "Go to Goodreads" command.

### Expected result:
Goodreads opens in a new tab.

### Actual result:
The tab doesn't go through the "Redirect Notice" to display the actual page and the doorhanger remains in the "One second..." state.

### Notes:
Reproduced on Mac 10.14 with Firefox Nightly 77.0a1 (2020-04-13).

dave-ok:

Fixes #1365 and Fixes #1526
@ianb Can you please review this? I also made an observation in #1365

### DIFF
--- a/extension/background/intentRunner.js
+++ b/extension/background/intentRunner.js
@@ -223,7 +223,9 @@ export class IntentContext {
   }
 
   async createTab(options) {
-    return browserUtil.createTab(options);
+    const tab = await browserUtil.createTab(options);
+    await browserUtil.loadUrl(tab.id, options.url);
+    return tab;
   }
 
   async openOrFocusTab(url) {
--- a/extension/background/intentRunner.js
+++ b/extension/background/intentRunner.js
@@ -223,9 +223,7 @@ export class IntentContext {
   }
 
   async createTab(options) {
-    const tab = await browserUtil.createTab(options);
-    await browserUtil.loadUrl(tab.id, options.url);
-    return tab;
+    return browserUtil.createTab(options);
   }
 
   async openOrFocusTab(url) {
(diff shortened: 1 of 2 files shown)


ianb:
Hmm... that `context.createTab()` and `browserUtil.createTab()` are different here is problematic, but I suspect that this will regress something else. This reverts https://github.com/mozilla/firefox-voice/commit/d01e67e9

But I'm not sure what the right resolution here is. I think there's probably something else (e.g., loadUrl) that needs to be fixed, or we have to distinguish between expected-final URLs (i.e., a destination) and expected-interstitial URLs (like an I'm Feeling Luck URL)

ianb:
Actually I think the bug here is in `loadUrl`, at least if my guess at the problem here is correct. If you load a URL, and that URL redirects, that should be a success. We only want it to wait as long as it takes to get _away_ from the old URL. Right now we require that the URL actually resolves to the expected URL.

dave-ok:
How about if we check if the URL is already the expected URL immediately in `loadUrl` and return right away if it is the expected URL. Then we don't have to wait anymore

dave-ok:
@ianb I have made some changes. Can you please confirm if this is an acceptable solution?

ianb:
While a better version of `loadUrl` would help here, really we shouldn't use `loadUrl` at all in the Google I'm Lucky searches. So I've created  #1589 that does that.

Actually fixing `loadUrl` would mean somehow deciding when to resolve. While it would be reasonable to detect the Google redirect in that function (even though it's a special case), in that case it shouldn't resolve until the redirect is forced through. Or alternately, we just detect when something has redirected and complete, perhaps by checking that the page has had at least one URL change, and that it has resolved to a status of "complete".

dave-ok:
Thank you for explaining this thoroughly. I had a feeling my solution was too particular and a bit hacky anyways :)

dave-ok:
@ianb I think #1589 should close #1365 as well?

PR #1671, opened 2020-05-21
Fix #1506 default value for type input text area

### LINKED ISSUE #1506: [object Object] present in "Type Input" when "Developer tip" link is accessed from doorhanger
### Prerequisites:
Mic permissions are enabled.

### STR:
1. Use the shortcut or click on the mic icon from the browser toolbar and wait a few seconds until the doorhanger extends.
2. Click the link under "Developer tip".

### Expected result:
"Type Input" doesn't have display issues.

### Actual result:
[object Object] is written in the "Type Input."

### Notes:
Reproduced on Mac 10.14 with Firefox Nightly 77.0a1 (2020-04-08).
Multiple tries may be needed for the doorhanger to successfully extend. Covering the mic may help.

Simpcyclassy:

### DIFF
--- a/extension/popup/popupView.jsx
+++ b/extension/popup/popupView.jsx
@@ -547,7 +547,10 @@ const VoiceInput = ({
           {buildSettings.inDevelopment ? (
             <div>
               <p>Developer tip:</p>
-              <button className="type-input-button" onClick={onInputStarted}>
+              <button
+                className="type-input-button"
+                onClick={() => onInputStarted("")}
+              >
                 Start typing to make a request using a keyboard
               </button>
             </div>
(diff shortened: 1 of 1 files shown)


extension/popup/popupView.jsx
```diff
@@ -684,7 +684,6 @@ class TypingInput extends PureComponent {
           autofocus="1"
           onKeyPress={this.onInputKeyPress}
           onChange={this.onInputTextChange}
-          defaultValue={this.value}
```
ianb, line 687: This is here so that the first keypress gets entered into the input field, so removing it will cause problems with typed input.

The problem is that this is how onInputStarted is intended to work:

```js
  onInputTextChange = event => {
    this.value = event.target.value;
    if (this.value && this.props.onInputStarted) {
      this.props.onInputStarted(this.value);
    }
  };
```

But when clicked it's used like this:

```js
<button className="type-input-button" onClick={onInputStarted}>
```

In that case `onInputStarted` is called with an event, which is an object. Instead it needs to be something like `onClick={() => onInputStarted("")}`

Simpcyclassy: Alright, it makes sense. Fixing that now.

PR #1674, opened 2020-05-21
Fix #1516, add FAQ entry about add-on permissions

### LINKED ISSUE #1516: Inform users about required permissions before installing the extension
A user in the chat room stated that it was a bit shocking seeing all the permissions that were requested on the first use of the extension. The suggestion they proffered was to link to the privacy policy on the page where a user installs the extension, so that a user can read it beforehand.

Not sure if others have had a similar experience, so I created this issue in their behalf to track it in case.

ianb:
New FAQ entry:

image

### DIFF
--- a/homepage/site/homepageView.jsx
+++ b/homepage/site/homepageView.jsx
@@ -442,6 +442,80 @@ const Faq = () => {
         <details>
           <summary>
             <div class="faq-section-header">
+              Why does Firefox Voice ask for all these permissions?
+            </div>
+          </summary>
+          <p>
+            Because Firefox Voice implements a wide variety of commands, it
+            needs permissions to do many things on your behalf.
+          </p>
+          <dl>
+            <dt>Access your data for all websites</dt>
+            <dd>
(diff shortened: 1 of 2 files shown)


homepage/site/homepageView.jsx
```diff
@@ -439,6 +439,80 @@ const Faq = () => {
...
+            <dt>Read the text of all open tabs</dt>
+            <dd>
+              This permission is used for the command "find [query] in page",
+              which is the equivalent of ^F/⌘F.
+            </dd>
+            <dt>Access browsing history</dt>
+            <dd>
+              This is used to determine your preferred music service, and in the
+              future we plan to use it to detect other preferred services. E.g.,
+              if you visit `spotify.com` often/recently, then we presume that is
```
ianb, line 502: I apparently forgot this isn't Markdown. I'll fix this along with what I'm guessing may be some incoming language changes.

ianb:
Committed with changes per the doc applied

PR #1686, opened 2020-05-27
Add moment-by-moment timing logs

ianb:
This extends #1684, adding logging for individual bits of actions


ianb:
Whoops, this included way too much...

PR #1688, opened 2020-05-27
Add email intents and services

ianb:
This adds an intent to compose an email, with an optional subject and to address, and optionally based on the current page.

Also the service framework is extended for email providers, and a gmail provider added. Not much functionality around those providers yet.


ianb:
Doh, overcommitted again!

PR #1701, opened 2020-05-28
Fixes #1609

### LINKED ISSUE #1609: Catch "Missing host permission for the tab"
In several cases we'll get the error "Missing host permission for the tab" – instead we should rewrite that phrase as "That does not work on this kind of page"

We should probably change pageMetadata to also work around this, and give partial metadata for these pages.

An example of getting this error is if you screenshot about:newtab – and an example where this shouldn't give an error (with a pageMetadata update) is "read page title"

danielamormocea:
Fixes #1609

### DIFF
--- a/extension/background/content.js
+++ b/extension/background/content.js
@@ -52,9 +52,17 @@ export async function lazyInject(tabId, scripts) {
   ]
     .concat(scripts)
     .concat(["/content/responder.js"]);
+
   for (const script of scripts) {
-    await browser.tabs.executeScript(tabId, { file: script });
+    try {
+      await browser.tabs.executeScript(tabId, { file: script });
+    } catch (error) {
+      const e = new Error("That does not work on this kind of page");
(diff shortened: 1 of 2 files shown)


extension/background/content.js
```diff
@@ -52,9 +52,17 @@ export async function lazyInject(tabId, scripts) {
...
+
   for (const script of scripts) {
-    await browser.tabs.executeScript(tabId, { file: script });
+    try {
+      await browser.tabs.executeScript(tabId, { file: script });
+    } catch (error) {
+      const e = new Error("That does not work on this kind of page");
+      e.displayMessage = "That does not work on this kind of page";
+      throw e;
+    }
```
ianb, line 63: This will catch everything, which we don't want. We only want to catch the missing host permission.

So this should be something like:

```js
} catch (error) {
  if (error.message.includes("missing host permission")) { // Not sure if this is the right string
    const e = ...
  } else {
    throw error;
  }
}
```

PR #1713, opened 2020-06-04
When alarm finishes play audio twice

Simpcyclassy:


Simpcyclassy:
@ian the doorhanger closes immediately sometimes even before the first alarm ends. I am not clear which part of the code handles how long the doorhanger stays open

PR #1714, opened 2020-06-05
Fixes #1672 added mic selector preference

### LINKED ISSUE #1672: Put mic selector in preferences
Using [enumerateDevices](https://developer.mozilla.org/en-US/docs/Web/API/MediaDevices/enumerateDevices) we can let people select their microphone without changing permissions.

It would be nice to have this in the options, near the top. A select box would be fine.

danielamormocea:
Fixes #1672

I added in preferences a dropdown with the other audio input devices available.

### DIFF
--- a/extension/options/options.css
+++ b/extension/options/options.css
@@ -428,3 +428,7 @@ a:hover {
   right: 15px;
   position: absolute;
 }
+
+#mic-selector{
+  margin-top:20px;
+}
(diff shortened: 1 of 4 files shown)


ianb:
@danielamormocea Looks good, can you make a followup bug to improve the styling?

@awallin here's the current style:
image

PR #1725, opened 2020-06-09
Fix #1677 send synthetic keys

### LINKED ISSUE #1677: Send synthetic keys
For several cases automation will be much easier if we can simply send real keyboard events into pages. Then we often can map commands directly to keyboard events without any other special trickery.

The best source I can find for this is in geckodriver/marionette: https://searchfox.org/mozilla-central/source/testing/marionette/event.js

Here's the more specific function: https://searchfox.org/mozilla-central/source/testing/marionette/event.js#399-430

There are some potential security issues here, it's possible to imagine firing a keyboard event that does something unexpected in just the right context. We should separate textual input keyboard events (that simulate typing into a field) from anything that can emit non-textual keyboard events (with modifier keys besides Shift, arrow keys, etc). Then we should never allow user input or input collected from a page to create non-content keyboard events.

Simpcyclassy:


Simpcyclassy:
I skipped removing my PRs from draft yesterday when I requested a review

ianb:
I've made a few changes and pushed them to this branch: https://github.com/mozilla/firefox-voice/tree/pr/1725

"Scroll down" still doesn't work, but this is slightly closer to working.

Simpcyclassy:
Will this PR be closed or do I replicate the changes here?

extension/experiments/voice/api.js
```diff
@@ -101,6 +733,28 @@ this.voice = class extends ExtensionAPI {
...
+                    ? "Cannot access activeTab"
+                    : `Invalid tab ID: ${tabId}`
+                );
+              }
+              return tab;
+            }
+
+            const tab = getTabOrActive(tabId);
+            console.log("tab is", tab, tab.KeyboardEvent);
+            const KeyboardEvent = new KeyboardEvent(eventName, eventProperties);
```
ianb, line 755: For eventProperties see https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/KeyboardEvent

ianb:
One trick to look at the real events:

```js
document.addEventListener("keydown", event => {window.E = event})
```

Then hit a key, and look at the `E` variable to see what exactly that keypress looks like as an event.

PR #1730, opened 2020-06-10
fixed mute/unmute commands

### LINKED ISSUE #1729: Mute/Unmute Commands not working perfectly
Steps to Reproduce:
1. Start any web player.
2. Say or write the "Mute"  command.
3. Use  "unmute music" or "unmute volume"  command for unmuting.
 We are not able to use (mute/unmute music) or (mute/unmute volume) or vice versa.
Expected result:
Player should be unmuted.

Actual result:
Player is still muted.

Ishakikani9117:
fixes  #1729

### DIFF
--- a/extension/intents/music/music.toml
+++ b/extension/intents/music/music.toml
@@ -173,8 +173,8 @@ expectParameters = {volumeLevel = "levelUp"}
 [music.mute]
 description = "mute music volume"
 match = """
-  mute (music | audio | video) (volume |)
-  mute (volume | the volume)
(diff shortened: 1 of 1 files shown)


ianb:
@danielamormocea You've been looking at some of these phrases – I'm a little worried this will create some phrase ambiguities between music and tab muting, does that seem to be the case to you to?

ianb:
I don't think this PR will make sense, it catches too many phrases. We'll have to be more clever about deciding what "mute" really means given the browsing context.

PR #1744, opened 2020-06-12
Demo branch

ianb:
Note this is a branch to use for a demo, we'll have to figure out how to land this later


extension/wakewordTraining/wakewordTraining.css
```diff
@@ -0,0 +1,51 @@
+@import url("../options/options.css"); /* temp hack to get some of the options page styling in */
```
ianb, line 1: I'm not sure when we'll want to land the actual specific wakeword support. We'll have to discuss that separately. That would include the trainer.

ianb:
I think @jcambre has pulled much of the interesting things out of here, and anyway the comments/code are still available, so closing this PR

Issue #1751, opened 2020-06-17
Close all tabs closes active tab only


Simpcyclassy:
### Prerequisites:
Mic permissions are enabled.
"Allow" or "Don't Allow" button from “Allow Firefox Voice to Collect Voice Transcriptions” pop-up is clicked.

### STR:
1. Say or type "close tabs" command, e.g "close all tabs"

### Expected result:
All open tabs get closed

### Actual result:
Only active tab closes

### Notes:
Reproduced on Mac 10.15.5

PR #1753, opened 2020-06-19
Fixed issue #1751

veecee424:
Close all tabs command successfully closes all tabs

### DIFF
--- a/extension/browserUtil.js
+++ b/extension/browserUtil.js
@@ -7,6 +7,10 @@ export async function activeTab() {
   }))[0];
 }
 
+export async function activeTabs() {
+  return browser.tabs.query({});
+}
+
 export async function makeTabActive(tab) {
   let tabId;
   if (typeof tab === "string" || typeof tab === "number") {
--- a/extension/browserUtil.js
+++ b/extension/browserUtil.js
@@ -8,7 +8,11 @@ export async function activeTab() {
 }
 
 export async function activeTabs() {
-  return browser.tabs.query({});
+  return browser.tabs.query({
+    active: false,
+    currentWindow: true,
+    pinned: false,
+  });
 }
 
 export async function makeTabActive(tab) {
--- a/extension/intents/tabs/tabs.js
+++ b/extension/intents/tabs/tabs.js
@@ -108,9 +108,7 @@ intentRunner.registerIntent({
   name: "tabs.closeAll",
   async run(context) {
     const activeTabs = await context.activeTabs();
-    for (const tab of activeTabs) {
-      await browser.tabs.remove(tab.id);
-    }
+    browser.tabs.remove(activeTabs.map(tab => tab.id));
     context.displayText("All Tabs closed");
   },
 });
--- a/extension/intents/tabs/tabs.js
+++ b/extension/intents/tabs/tabs.js
@@ -105,6 +105,17 @@ intentRunner.registerIntent({
 });
 
 intentRunner.registerIntent({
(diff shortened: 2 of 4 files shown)


extension/browserUtil.js
```diff
@@ -7,6 +7,10 @@ export async function activeTab() {
   }))[0];
 }
 
+export async function activeTabs() {
```
ianb, line 10: I don't think this function does as advertised: it returns all tabs, not just the active ones.

It would be fine instead to run `browser.tabs.query()` inside the closeAll intent, with whatever options are appropriate there.

extension/intents/tabs/tabs.js
```diff
@@ -104,6 +104,17 @@ intentRunner.registerIntent({
   },
 });
 
+intentRunner.registerIntent({
+  name: "tabs.closeAll",
+  async run(context) {
+    const activeTabs = await context.activeTabs();
+    for (const tab of activeTabs) {
+      await browser.tabs.remove(tab.id);
```
ianb, line 112: You could actually remove all of them at once: `browser.tabs.remove(tabs.map(tab => tab.id))`

But closing all tabs wouldn't be right, it leaves Firefox in a weird state. I think it quits Firefox on Windows (maybe?) and on Mac it leaves Firefox open but with no windows.

Instead only tabs that aren't pinned should be closed, and only tabs in the current window. And maybe the active tab should also be left open. So that would be something like `browser.tabs.query({active: false, lastFocusedWindow: true, pinned: false})`

extension/intents/tabs/tabs.toml
```diff
@@ -1,27 +1,23 @@
 [tabs.close]
 description = "Closes the current tab"
 match = """
-  close (the | this |) (active | current |) (tab{s} | site | page) (for me |)
+  close (the | this |) (active | current |) (tab | site | page) (for me |)
```
ianb, line 4: `tab{s}` should be left here – voice recognition often adds or removes plurals in cases like this, so we have to be fuzzy about how we match.

extension/intents/tabs/tabs.toml
```diff
@@ -454,3 +450,36 @@ expectParameters = { dir = "first" }
 [[tabs.selectNumbersTabs.example]]
 phrase = "select last two tabs"
 expectParameters = { dir = "last" }
+
+[tabs.closeAll]
+description = "Closes all active or open tabs"
+match = """
+  close (all | every | each | the |) (active | open |) (tab{s} | site{s} | page{s}) (for me |)
```
ianb, line 457: This should only match "close all tabs" or maybe "close every tab". So something like `close (all | every) tab{s} (for me|)`

extension/browserUtil.js
veecee424: done!

extension/intents/tabs/tabs.js
veecee424: done!

extension/intents/tabs/tabs.toml
veecee424: done!

extension/intents/tabs/tabs.toml
veecee424: done!

veecee424:
I erroneously was working off my master branch and got a few things messed up so I had to create a new branch hence a new PR effecting all the changes you requested. Sorry for the inconvenience.

ianb:
Closing in preference of  #1772

PR #1754, opened 2020-06-20
intent for close all tabs

### LINKED ISSUE #1751: Close all tabs closes active tab only

Ishakikani9117:
fixes #1751

### DIFF
--- a/extension/intents/tabs/tabs.js
+++ b/extension/intents/tabs/tabs.js
@@ -107,11 +107,12 @@ intentRunner.registerIntent({
 intentRunner.registerIntent({
   name: "tabs.closeAll",
   async run(context) {
-    while (1) {
-      const activeTab = await context.activeTab();
-      if (activeTab.id === 1) break;
-      await browser.tabs.remove(activeTab.id);
-    }
+    const openTabs = await browser.tabs.query({
+      active: false,
+      lastFocusedWindow: true,
+      pinned: false,
+    });
+    await browser.tabs.remove(openTabs.map(tab => tab.id));
     context.displayText("Closed all tabs");
   },
 });
--- a/extension/intents/tabs/tabs.js
+++ b/extension/intents/tabs/tabs.js
@@ -105,6 +105,18 @@ intentRunner.registerIntent({
 });
 
 intentRunner.registerIntent({
+  name: "tabs.closeAll",
+  async run(context) {
+    while (1) {
+      const activeTab = await context.activeTab();
+      if (activeTab.id === 1) break;
+      await browser.tabs.remove(activeTab.id);
+    }
+    context.displayText("Tab closed");
+  },
+});
+
+intentRunner.registerIntent({
   name: "tabs.undoCloseTab",
   async run(context) {
     await browser.experiments.voice.undoCloseTab();
--- a/extension/intents/tabs/tabs.js
+++ b/extension/intents/tabs/tabs.js
@@ -112,7 +112,7 @@ intentRunner.registerIntent({
       if (activeTab.id === 1) break;
       await browser.tabs.remove(activeTab.id);
     }
-    context.displayText("Tab closed");
+    context.displayText("Closed all tabs");
   },
 });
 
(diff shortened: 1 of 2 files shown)


Simpcyclassy:
I think this ticket is being worked on [here](https://github.com/mozilla/firefox-voice/pull/1753) also

Ishakikani9117:
@Simpcyclassy  Yes, you're right. But I think that I have a better solution.

extension/intents/tabs/tabs.js
```diff
@@ -104,6 +104,18 @@ intentRunner.registerIntent({
   },
 });
 
+intentRunner.registerIntent({
+  name: "tabs.closeAll",
+  async run(context) {
+    while (1) {
+      const activeTab = await context.activeTab();
+      if (activeTab.id === 1) break;
```
ianb, line 112: I don't understand quite what's happening here? It seems like it's going to close the active tab until it has that particular id. But the id should be stable, and there's no reason there will necessarily be an id of 1. Like if you open the browser, open a few tabs. Close a few of those first tabs, then there will be no tab with id 1.

Also this will act on tabs in other windows.

Since closing _all_ tabs doesn't work well, we should either open a new tab or else leave the active tab open, or close all but pinned tabs. 

So a query that matches all tabs that should be closed might be: `browser.tabs.query({active: false, lastFocusedWindow: true, pinned: false})`

extension/intents/tabs/tabs.toml
```diff
@@ -1,31 +1,45 @@
...
-phrase = "close active tab"
+phrase = "close the current top"
 test = true
 
-[[tabs.close.example]]
-phrase = "close the active tabs"
+[tabs.closeAll]
+description = "Closes all open tabs"
+match = """
+  close (all |) (active | open |) (tab{s} | site{s} | page{s}) (for me |)
```
ianb, line 26: "all" here should be required. `(all |)` makes the word optional

extension/intents/tabs/tabs.toml
```diff
@@ -1,31 +1,45 @@
...
+"""
+
+[[tabs.closeAll.example]]
+phrase = "close tabs"
```
ianb, line 30: This is an example where "all" should be required, something like "close tab" (which is also matched for the phrase) should only close the active tab, definitely not all tabs!

extension/intents/tabs/tabs.js
Ishakikani9117: @ianb I know that the id is not stable but for the first that opens up when you open up the browser always has id 1.
no matter how many times you open other tabs and close them. you can check that by using "console.log(activeTab.id);" . so my idea is to keep open that first tab with id 1.

extension/intents/tabs/tabs.toml
Ishakikani9117: done!

extension/intents/tabs/tabs.toml
Ishakikani9117: done!

extension/intents/tabs/tabs.js
Ishakikani9117: browser.tabs.query({active: false, lastFocusedWindow: true, pinned: false})  I tried this query in closeAll intent 
like : const openTabs = browser.tabs.query({active: false, lastFocusedWindow: true, pinned: false})
but it doesn't work. The intent runs but doesn't  close all tabs.

Ishakikani9117: @ianb  I have fixed it as you suggested.

ianb:
This was superceded by #1772 but your changes do look good. I'll land and revert this so there's a record of the contribution, even if it's duplicated elsewhere.

PR #1756, opened 2020-06-23
Fixes #1483 Create registerPageName function in intentRunner

### LINKED ISSUE #1483: Create registerPageName function in intentRunner
This is a followup to #1302 (per this comment: https://github.com/mozilla/firefox-voice/pull/1302#discussion_r396027674)

Right now `navigation.js` is directly accessing `browser.storage`, but instead it should call functions in `intentRunner`. These new functions would use the same pattern as `intentRunner.registerNickname()`

Simpcyclassy:


extension/background/intentRunner.js
```diff
@@ -531,6 +531,42 @@ export function getRegisteredNicknames() {
   return registeredNicknames;
 }
 
+const registeredPageName = {};
+
+export function registerPageName(name, context) {
```
ianb, line 536: Instead of context this should take `{url, creationDate}` (or maybe fill in creationDate itself)

extension/background/intentRunner.js
```diff
@@ -531,6 +531,42 @@ export function getRegisteredNicknames() {
...
+    for (const name in result.registeredPageName) {
+      const value = result.registeredPageName[name];
+      const context = new IntentContext(value);
+      registeredPageName[name] = context;
+      log.info("Loaded page name", name, context.name, context.slots);
+    }
+  }
+}
+
+export function getRegisteredPageName() {
```
ianb, line 566: This should be like `getRegisteredPageName(name)` – and return null if there's no page with that name.

extension/background/intentRunner.js
```diff
@@ -531,6 +531,42 @@ export function getRegisteredNicknames() {
...
 
+const registeredPageName = {};
+
+export function registerPageName(name, context) {
+  name = name.toLowerCase();
+  if (!context) {
+    delete registeredPageName[name];
+    log.info("Removed nickname for page", name);
+  } else {
+    registeredPageName[name] = context;
```
ianb, line 542: `name` should probably be lower-cased here (i.e., `name.toLowerCase()`)

Simpcyclassy: It is already lower-cased on line 551

Simpcyclassy:
I still get  a type error for remove page name "Internal error: TypeError: (destructured parameter) is null"

extension/intents/nicknames/nicknames.js
```diff
@@ -99,42 +99,18 @@ intentRunner.registerIntent({
...
-      const exc = new Error("No page name to remove");
-      exc.displayMessage = `No page name "${name}" found`;
-      throw exc;
-    }
-    delete pageNames[name];
-    log.info("Removed page name", name);
-    await browser.storage.sync.set({ pageNames });
+    const name = context.slots.name;
+    await intentRunner.getRegisteredPageName(name);
+    intentRunner.registerPageName(name, null);
```
ianb, line 114: Here's the error: this passes the second argument as `null`, but the registerPageName function expects an object with a property `url`. So it would work with `intentRunner.registerPageName(name, {url: null})`

Probably the best fix is to update `registerPageName` though so it can accept null as a second argument.

Simpcyclassy: Throws a `TypeError: can't access property "url", pagedata is null`

```
export async function registerPageName(name, pagedata) {
  name = name.toLowerCase();
  const creationDate = Date.now();
  const result = await browser.storage.sync.get("pageNames");
  pageNames = result.pageNames || {};
  const url = pagedata.url;
  pageNames[name] = url;

  if (!url) {
    delete pageNames[name];
    log.info("Removed nickname for page", name);
  } else {
    pageNames[name] = url;
    log.info("Added nickname for page", name, "->", url, creationDate);
  }
  await browser.storage.sync.set({ pageNames });
}
```

Simpcyclassy: But then when I do it this way `(name, {url: null})` works fine

extension/intents/nicknames/nicknames.js
```diff
@@ -99,42 +99,18 @@ intentRunner.registerIntent({
...
-      const exc = new Error("No page name to remove");
-      exc.displayMessage = `No page name "${name}" found`;
-      throw exc;
-    }
-    delete pageNames[name];
-    log.info("Removed page name", name);
-    await browser.storage.sync.set({ pageNames });
+    const name = context.slots.name;
+    await intentRunner.getRegisteredPageName(name);
+    intentRunner.registerPageName(name, { url: null });
```
ianb, line 114: I guess this works, but it doesn't make it clear what's happening. A null URL doesn't really imply the page will be unregistered.

I guess it would be best to just create another function `intentRunner.unregisterPageName(name)`

PR #1762, opened 2020-06-24
Fixes #963 added intents for missing features

### LINKED ISSUE #963: Put in stub intent to catch known missing features
We have missing features like #919 that are very clear. Having these open up a search page is not helpful.

We could make one big bucket "we know what you wanted to do, but we aren't doing it" intent that captures these.

We should be sure this shows up well in Telemetry, and that it has feedback buttons.

While this seems somewhat involved, I think it's actually reasonably simple. We'll create a catch-all intent, like `self.notImplemented` and it will have phrases like:

```
match = """
set alarm [subtype=alarm] [message=Sorry, Firefox Voice doesn't include alarms yet]
restart Firefox [subtype=restart] [message=blah blah]
"""
```

There will be a default message, but it can be overridden. The `subtype` parameter should become a general feature (handled in `intentRunner` I think) that will change Telemetry intentName to, for instance, `self.notImplemented.alarm`

Actually collecting these phrases is a task for just reading through the submitted utterances, filtering for fallbacks, and picking out phrases that are clear but not implemented.

danielamormocea:
Fixes #963 

Added fallback.js and fallback.toml to catch the fallback intents and provide a custom message for these.


danielamormocea:
Don't merge this yet as I'm adding close and refresh functionalities

extension/background/intentParser.js
```diff
@@ -6,6 +6,7 @@ import { entityTypes } from "./entityTypes.js";
 
 const DEFAULT_INTENT = "search.search";
 const DEFAULT_SLOT = "query";
+const FALLBACK_INTENT = "fallbacks.notImplemented";
```
ianb, line 9: It doesn't look like this is used for anything...?

PR #1772, opened 2020-06-26
fixes #1751 close all tabs intent

### LINKED ISSUE #1751: Close all tabs closes active tab only

veecee424:
This is the new PR.

### DIFF
--- a/extension/browserUtil.js
+++ b/extension/browserUtil.js
@@ -8,6 +8,14 @@ export async function activeTab() {
   }))[0];
 }
 
+export async function activeTabs() {
+  return browser.tabs.query({
+    active: false,
+    currentWindow: true,
+    pinned: false,
+  });
+}
+
 export async function makeTabActive(tab) {
   let tabId;
   if (typeof tab === "string" || typeof tab === "number") {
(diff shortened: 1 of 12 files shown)


extension/browserUtil.js
```diff
@@ -8,6 +8,14 @@ export async function activeTab() {
   }))[0];
 }
 
+export async function activeTabs() {
```
ianb, line 11: This name doesn't make sense for the function – the function returns all tabs except the active/pinned tabs. But I don't think we need a function for this anyway, it can just go directly in the intent.

extension/intents/tabs/tabs.js
```diff
@@ -108,6 +108,15 @@ intentRunner.registerIntent({
   },
 });
 
+intentRunner.registerIntent({
+  name: "tabs.closeAll",
+  async run(context) {
+    const activeTabs = await context.activeTabs();
```
ianb, line 114: Instead of creating a function, you could just put the `browser.tabs.query(...)` call right here.

ianb:
It's looking good, thanks @veecee424 !

PR #1788, opened 2020-06-30
Fixes #1560

### LINKED ISSUE #1560: 'Sign in' and 'sign out' intent
Sign in and sign out are common links/functionality on many websites. To make the process more efficient Firefox Voice could click the link for a user.

When a user says `Sign in` (or login, log me in, sign me in) Firefox Voice should look for a matching link/button on the page and click it taking the user to the appropriate page/form.

When a user says `Sign out` (or log out, log me out, sign me out) Firefox Voice should look for a matching link/button on the page and click it.

If these the appropriate link can't be found a message saying so should be displayed (see attached).

sign-in-not-found

Simpcyclassy:
- [x] Find sign in and sign out slots
- [x] Click on the link
- [x] Render a unique FollowupContainer container when the mic is on, on a search result to handle the generic "waiting" text


extension/intents/tabs/tabs.toml
```diff
@@ -401,14 +401,23 @@ description = "Find on page content"
 match = """
   find [query] on (this |) (page | tab)
   search (this |) (page | tab) for [query]
+  [query]
```
ianb, line 404: This should be a new intent. Also we definitely don't want a match that catches everything!

The implementation is going to look a lot like the `navigation.followLink` intent (and can probably share code with it). As proposed "sign in" will just click a link matching `sign in | login | log in` (note unlike followLink it doesn't matter as much if the link is visible on the screen – followLink prefers links that are visible on the screen)

extension/intents/navigation/clickLink.js
```diff
@@ -0,0 +1,70 @@
...
+    };
+    const combinedContent = [];
+    for (const link of findLinks()) {
+      combinedContent.push({
+        element: link,
+        text: link.innerText,
+        title: link.getAttribute("title"),
+        url: link.url,
+      });
+    }
```
ianb, line 41: Instead of using Fuse, you could probably just test for something like `/log\s*in|sign\s*in/i.test(string)`, since sites will only use very limited strings for these buttons.

extension/intents/navigation/clickLink.js
```diff
@@ -0,0 +1,70 @@
...
+    let found;
+    for (const match of matches) {
+      const el = match.item.element;
+      return (found = el);
+    }
+    if (!found) {
+      found = matches[0].item.element;
+    }
+    highlightElement(found);
```
ianb, line 55: Here you should also scroll it into view, like `found.scrollIntoView()`

extension/intents/navigation/clickLink.js
```diff
@@ -0,0 +1,70 @@
...
+    }
+    const fuse = new Fuse(combinedContent, options);
+    const matches = fuse.search(query);
+    if (!matches.length) {
+      return false;
+    }
+    let found;
+    for (const match of matches) {
+      const el = match.item.element;
+      return (found = el);
```
ianb, line 50: The return here is definitely not correct. Try to keep this function and the loops simpler

extension/intents/navigation/navigation.toml
```diff
@@ -146,6 +146,45 @@ match = """
...
 """
 
+[[navigation.followLink.example]]
+phrase = "follow link latest news"
+
+
+[navigation.signInAndOut]
+description = "Follow the named sign in or out link and click"
+match = """
+  (sign in | login | log in)
```
ianb, line 156: If you have two different phrases, you'll need to do:

```
(sign in | login | log in) [action=login]
(sign out | logout | log out) [action=logout]
```

Then in the code:

```
run(context) {
  if (context.parameters.action === "login") {
    //login
  } else {
    //logout
  }
```

extension/intents/navigation/navigation.toml
```diff
@@ -146,6 +146,45 @@ match = """
...
+  (sign out | logout | log out)
+"""
+
+[[navigation.signInAndOut.example]]
+phrase = "sign in"
+expectSlots = {query = "sign in"}
```
ianb, line 162: this will fail, since there are no slots

extension/intents/navigation/clickLogout.js
```diff
@@ -0,0 +1,31 @@
...
+    const found = matches[0];
+
+    highlightElement(found);
+    found.scrollIntoView();
+    setTimeout(() => {
+      log.info("Following link to:", found.href || "?");
+      found.click();
+    }, 100);
+    return true;
+  });
```
ianb, line 22: These two (signIn and signOut) could just be in the same file

extension/intents/navigation/clickLogin.js
```diff
@@ -0,0 +1,31 @@
...
+  communicate.register("signIn", message => {
+    const { query } = message;
+    const regex = /log\s*in|sign\s*in/i;
+
+    for (const link of findButton()) {
+      const element = link.innerText;
+      return regex.test(element.innerText);
+    }
+    const matches = regex.match(query);
+    const found = matches[0];
```
ianb, line 13: This still has the same problem as before, and I don't think it will work at all. Be sure to test this!

The `return` statement will terminate the function as soon as you reach it! That means everything below the `for` loop will be skipped.

Also the next statements getting matches and assigning `found` don't really make sense.

extension/intents/navigation/navigation.js
```diff
@@ -191,6 +191,38 @@ intentRunner.registerIntent({
...
+      "/intents/navigation/clickLogout.js",
+    ]);
+    if (context.parameters.action === "login") {
+      const found = await browser.tabs.sendMessage(activeTab.id, {
+        type: "signIn",
+        query: context.slots.query,
+      });
+      if (found === false) {
+        const exc = new Error("No link found matching query");
+        exc.displayMessage = `Sorry can't "${context.slots.query}" on this page`;
```
ianb, line 209: There's no query slot on these phrases. You could use `context.utterance` here I believe. Please make sure to test on a page that has no sign in!

extension/intents/navigation/clickLogin.js
Simpcyclassy: Yeeeaaah. My bad I did not commit correctly

extension/intents/navigation/navigation.js
Simpcyclassy: Alright

extension/popup/popupView.jsx
```diff
@@ -902,7 +904,32 @@ const SearchResultsContent = ({
...
+                  id="search-show-next"
+                  onClick={onNextResultClick}
+                >
+                  {new URL(next.url).hostname} | {next.title}
+                </a>
+              </div>
+            ) : null}
+          </div>
+        </React.Fragment>
+      ) : (
```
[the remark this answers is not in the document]
Simpcyclassy: I will remove it. And add it where appriopriate. Was supposed to be part of the music service fix or I can create a new issue entirely for it.

PR #1791, opened 2020-07-05
Routines

### LINKED ISSUE #1758: Check routine start/stop times
Routines let us execute several actions in a row. This means intents are run one after another in quick sequence, without a user determining when the "right" moment is to execute an action.

For some ideas of possible routines, [see this brainstorming document](https://coda.io/d/Routines-in-Voice_dsSFX9HoxlM/Use-case-expansion_sud0N#Phrases_tuXj_/r5)

Right now many intents will fail when run in sequence, because the intent finishes too quickly, _or_ starts to quickly. For instance, an intent may open a page, but it completes before the page is really open. Or an intent may try to interact with a page, but it interacts with the page before it is fully loaded and responsive.

One example is a possible routine "paste screenshot on Twitter":

```
copy screenshot
open twitter
paste
```

The common problem is that "paste" doesn't work right away. Twitter will appear to load, but it fills in extra JavaScript later. To fix this probably requires "paste" to retry if it appears not to work.

danielamormocea:
Fixes #1758 

I added a content script to check with the help of selectors or with default "html" if a page is fully loaded. The default should behave similarly to "waitForDocumentComplete". I figured the checks on navigation and search should be enough for the moment. In general, I think google sites (as gmail - already fixed) would require to visually see what's the last thing that's loading and specify that manually.

### DIFF
--- a/extension/browserUtil.js
+++ b/extension/browserUtil.js
@@ -1,7 +1,6 @@
 /* globals buildSettings */
 import * as searching from "./searching.js";
 import * as content from "./background/content.js";
-const PAGE_LOAD_CHECKER_SCRIPT = "./content/pageLoadChecker.js";
 
 export async function activeTab() {
   return (await browser.tabs.query({
@@ -225,7 +224,7 @@ export function waitForDocumentComplete(tabId) {
 }
 
 export async function waitForPageToLoadUsingSelector(tabId, options = {}) {
-  await content.inject(tabId, PAGE_LOAD_CHECKER_SCRIPT);
+  await content.inject(tabId, "./content/pageLoadChecker.js");
   return browser.tabs.sendMessage(tabId, { type: "isLoaded", options });
 }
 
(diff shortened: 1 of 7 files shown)


extension/browserUtil.js
```diff
@@ -1,5 +1,7 @@
 /* globals buildSettings */
 import * as searching from "./searching.js";
+import * as content from "./background/content.js";
+const PAGE_LOAD_CHECKER_SCRIPT = "./content/pageLoadChecker.js";
```
ianb, line 4: If something is only used once, it probably doesn't need to be a constant, you can just put the string in where it is used

extension/content/pageLoadChecker.js
```diff
@@ -0,0 +1,26 @@
...
+        if (document.querySelector(options.selector || DEFAULT_SELECTOR)) {
+          resolve(true);
+          clearInterval(checkExist);
+          clearTimeout(timeout);
+        }
+      }, options.checkInterval || DEFAULT_CHECK_INTERVAL);
+      // the specified selector has not been found
+      timeout = setTimeout(() => {
+        reject(new Error("The page took too much to load."));
+        clearInterval(checkExist);
```
ianb, line 22: This should probably `resolve(false)`

PR #1795, opened 2020-07-07
Count all tabs in all active browser windows

### LINKED ISSUE #1776: "How many tabs do I have open" should report all tabs, not just the ones in the current window
Right now, "How many tabs do I have open" reports one number, the number of tabs in the current window. 

We should report the number of tabs in _all_ open Firefox windows ("You have 33 tabs in 4 windows"); optionally we could additionally report the number of tabs in the local window ("You have 33 tabs in 4 windows, of which 8 are in this window.")

PascalUlor:
This returns a count of all tabs in all active browser windows. It also returns a count of the browser windows open.
Fixes  #1776

zoom-0


Simpcyclassy:
Please enable your lint, looks like you have a couple of linting and indentation issues on `/extension/intents/tabs/tabs.js`

extension/background/language/langs/english.js
```diff
@@ -0,0 +1,343 @@
+// Generated from english.toml
```
ianb, line 1: This file is generated, and shouldn't be added to the commit

extension/popup/popup.css
```diff
@@ -196,7 +196,7 @@ body,
 }
 
 .results-largeText {
-  font-size: 72px;
+  font-size: 1em;
```
ianb, line 199: I don't think we want to change this. Instead we should use something else (besides `largeText`) above

extension/intents/tabs/tabs.js
```diff
@@ -606,10 +606,20 @@ intentRunner.registerIntent({
...
+    let windowsCount = 1;
+    sampleTabs.forEach((window, index) => {
+      tabsCount += window.tabs.length ;
+      windowsCount += index;
+    });
 
     const card = {
       answer: {
-        largeText: `${numOfOpenTabs}`,
+        largeText: `Open Windows: ${windowsCount} ,
```
ianb, line 619: I think if you switch this to `text:` instead of `largeText:` that this will display better.

PascalUlor: The component rendering answers is referencing `largeText` as output
```
 <div className="results-text">
        <em>
          {card.answer.largeText ? (
            <div className="results-largeText">{card.answer.largeText}</div>
          ) : null}
          <div>{card.answer.text}</div>
        </em>
      </div>
```
I could change the CSS class to `results-text`

PascalUlor:
I've implemented the changes requested @ianb and @Simpcyclassy

PascalUlor:
@ianb I've implemented your feedback and I'm awaiting further instructions if any

extension/intents/tabs/tabs.js
PascalUlor: Since `largeText` is being used by other outputs I created a function to handle the text based on which tab method is called

PascalUlor:
@ianb I've implemented your feedback and I'm awaiting further instructions if any

ianb:
This is getting a little complicated. I don't really want to change the styles too much, but I'm going to land this and make some changes

PR #1796, opened 2020-07-07
Fixes  #1268 [Intent] What music services do you support?

### LINKED ISSUE #1268: [Intent] What music services do you support?
People are asking what music services are supported. The results should be displayed in a card with the ability to set a default. See mockup for details. Of note:

* Music services support will be growing over time so the query should current the current status at the time of request.
* Clicking on an icon/name should set the default music service. A hint to this affect is displayed at the bottom of the card.
* Attaching icons for the major services (even if we don't support them yet)

[music service.zip](https://github.com/mozilla/firefox-voice/files/4326752/music.service.zip)

music-provider

Simpcyclassy:


extension/popup/popupView.jsx
```diff
@@ -890,11 +912,40 @@ const SearchResultsContent = ({
...
+                  <img
+                    className="results-image music-service-image"
+                    src={imgSrc}
+                    alt={alt}
+                  />
+                </div>
+                <div className="results-medium-text">{text}</div>
+              </option>
+            ))}
+        </select>
```
Simpcyclassy, line 937: I am not sure how to handle the options here @ianb , musicServices appear to be an array.

extension/popup/popupController.jsx
```diff
@@ -169,6 +172,15 @@ export const PopupController = function() {
     });
     window.close();
   };
+  const initSettings = async () => {
+    const result = await settings.getSettingsAndOptions();
+    setUserSettings(result.settings);
+    setUserOptions(result.options);
+  };
```
ianb, line 179: We don't want to add more initialization here, especially since it will almost never be needed - only if you use this one intent will this be necessary. If you look in `historyView.jsx:History()` you'll see an example of using an effect inside the component to fetch data. In the same way you could fetch this data inside the `MusicCard` component, and then it would only be fetched when the card was displayed.

extension/popup/popupView.jsx
```diff
@@ -890,11 +912,34 @@ const SearchResultsContent = ({
     </div>
   );
 
+  const MusicCard = () => (
```
ianb, line 915: This should be called MusicSettingCard, since it's more specific than just music.

extension/popup/popupView.jsx
```diff
@@ -863,6 +885,13 @@ const SearchResultsContent = ({
     onNextSearchResultClick();
   };
 
+  const onMusicServiceChange = event => {
+    if (event) {
```
ianb, line 889: Is there any reason why the event would be false/null?

extension/popup/popupView.jsx
```diff
@@ -890,11 +919,36 @@ const SearchResultsContent = ({
...
+    <div value={userSettings.musicService} className="results-set">
+      {userOptions.musicServices &&
+        userOptions.musicServices
+          .filter(({ name }) => name !== "auto")
+          .map(({ name, title, imgSrc }) => (
+            <button
+              key={name}
+              value={name}
+              onChange={() => {
+                onMusicServiceChange(name);
```
ianb, line 932: Here you pass in the name in, which would be OK too, but that doesn't match the code above, which uses `event.target.value`.

I don't think this code will work as written?

extension/services/deezer/deezer.js
```diff
@@ -7,6 +7,7 @@ Object.assign(Deezer, {
   id: "deezer",
   title: "Deezer",
   baseUrl: "https://www.deezer.com/",
+  imgSrc: "../../assets/images/deezer.svg",
```
ianb, line 10: Since this could be used anywhere (e.g., in the popup), it's not clear what the relative path is relative to. Instead it should just be an absolute path (`/assets/images/deezer.svg`)

extension/intents/music/music.toml
```diff
@@ -240,3 +240,13 @@ match = """
 
 [[music.playPlaylist.example]]
 phrase = "play discover weekly playlist on spotify"
+
+
+[music.support]
+description = "This shows the results of music services supported in a card with the ability to set a default"
+match = """
+  What (are your|) (music | streaming |) (service{s} | player{s} | radio) (do you|) (support | have |)
```
ianb, line 248: `(what | which)` – both words seem reasonable.

extension/popup/popupView.jsx
Simpcyclassy: I really do not think so cause before this is even called an event is being triggered

Simpcyclassy:
image
I tried to implement what I did previously where I was initializing in `popuController` and for some weird reason it was not working so I went back in time to my previous commits and implemented and fetched data only when needed and the exact implementation on the `useEffect` that did not work now works perfectly fine, even more weird, causes the changes I made were same.
So, now it fetches the data but when I click on any of the services it does not seem to change the default service as desired

Could you take a look to see what I am doing wrong @ianb

extension/popup/popupController.jsx
```diff
@@ -91,14 +92,15 @@ export const PopupController = function() {
   const init = async () => {
     log.timing("PopupController init() called");
     const userSettings = await settings.getSettings();
+    // await initSettings();
```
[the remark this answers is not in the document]
Simpcyclassy: My bad

PR #1798, opened 2020-07-07
Partly fixes #1757 pause and continue routines

### LINKED ISSUE #1757: Pause/continue in routines
There are several cases when we might want to run some parts of a routine, then the routine should pause while the user executes some action (e.g., labels an issue, composes an email, etc)

You could imagine a Triage routine:

```
go to https://github.com/mozilla/firefox-voice/issues?q=is%3Aopen+is%3Aissue+no%3Amilestone
copy all links like https://github.com/mozilla/firefox-voice/issues/
for each link in clipboard
open link
pause
next link
```

When you hit `pause` it should stop, and the command `continue` would continue to execute the routine.

While "copy all links" and "for each" don't exist, this also is to demonstrate that the pause/continue may require keeping more context than a simple list index.

danielamormocea:
WIP #1757


extension/intents/nicknames/nicknames.js
```diff
@@ -114,3 +102,43 @@ intentRunner.registerIntent({
...
+intentRunner.registerIntent({
+  name: "nicknames.pause",
+  async run(context) {
+    if (context.parentRoutine === undefined) {
+      const exc = new Error("Command not available");
+      exc.displayMessage = "Command not available";
+      throw exc;
+    }
+    context.parentRoutine.stop = true;
+    browser.storage.sync.set({
```
ianb, line 115: Hmm... this would keep it over restarts, but I'm not sure if that's necessary. But at least it should use `browser.storage.local` as it wouldn't make sense to restart from another computer/profile (so it doesn't need to be sync'd).

extension/intents/nicknames/nicknames.js
```diff
@@ -114,3 +102,43 @@ intentRunner.registerIntent({
...
+  },
+});
+
+intentRunner.registerIntent({
+  name: "nicknames.continue",
+  async run(context) {
+    const { pausedRoutine } = await browser.storage.sync.get("pausedRoutine");
+    if (pausedRoutine === undefined) {
+      const exc = new Error("Command not available");
+      exc.displayMessage = "Command not available";
```
ianb, line 130: This should have a clearer message. Maybe like `continue can only be used in a routine after "pause routine"`

extension/intents/nicknames/nicknames.js
```diff
@@ -114,3 +102,43 @@ intentRunner.registerIntent({
...
+
+intentRunner.registerIntent({
+  name: "nicknames.pause",
+  async run(context) {
+    if (context.parentRoutine === undefined) {
+      const exc = new Error("Command not available");
+      exc.displayMessage = "Command not available";
+      throw exc;
+    }
+    context.parentRoutine.stop = true;
```
ianb, line 114: `.stop` should be a function instead of a flag. (It might just set `this._stop = true`, but from the outside it should be a function)

extension/intents/nicknames/nicknames.toml
```diff
@@ -60,4 +60,22 @@ match = """
...
+phrase = "remove this tab name"
+
+[nicknames.pause]
+description = "Pauses a routine"
+match = """
+  pause routine
+"""
+
+[[nicknames.pause.example]]
+phrase = "pause routine"
```
ianb, line 72: There doesn't need to be an example here. If we describe it somewhere, it should be on the routine page itself.

extension/intents/nicknames/nicknames.toml
```diff
@@ -60,4 +60,22 @@ match = """
...
+
+[nicknames.continue]
+description = "Continues a routine"
+match = """
+  continue routine
+"""
+
+[[nicknames.continue.example]]
+phrase = "Continue routine"
```
ianb, line 81: Similarly this doesn't need an example

extension/intents/nicknames/nicknames.toml
```diff
@@ -60,4 +60,22 @@ match = """
...
 
 [[nicknames.removePageName.example]]
-phrase = "remove this tab name"
\ No newline at end of file
+phrase = "remove this tab name"
+
+[nicknames.pause]
+description = "Pauses a routine"
+match = """
+  pause routine
```
ianb, line 68: For a message this could also have a phrase like `pause routine (with |) (and |) (show |) (message | status | text | print) [message]`

extension/intents/nicknames/routineExecutor.js
```diff
@@ -0,0 +1,48 @@
...
+    let hadError = false;
+    subcommand.onError = () => {
+      hadError = true;
+    };
+
+    subcommand.parentRoutine = {
+      name: this.routineName,
+      nextIndex: this.index + 1,
+      stop: false,
+    };
```
ianb, line 22: I think instead I would do something like `subcommand.routineExecutor = this` and then pause can access methods on this object.

extension/intents/nicknames/routineExecutor.js
```diff
@@ -0,0 +1,48 @@
...
+  constructor(routineName, subcommands, index = 0) {
+    this.routineName = routineName;
+    this.subcommands = subcommands;
+    this.index = index;
+  }
+
+  async runSubcommand(subcommand) {
+    let hadError = false;
+    subcommand.onError = () => {
+      hadError = true;
```
ianb, line 15: This should capture the message too, like `subcommand.onError = (message) => {hadError=true; errorMessage=message}`

extension/intents/nicknames/routineExecutor.js
```diff
@@ -0,0 +1,48 @@
...
+      "  Running subintent",
+      subcommand,
+      subcommand.name,
+      subcommand.slots
+    );
+
+    await intentRunner.runIntent(subcommand);
+    if (hadError) {
+      log.info("  Last intent failed, stopping");
+      return true;
```
ianb, line 33: And then this should probably show the error message and the routine name (i.e., `routine {name} encountered error {error} while running {subcommand.utterance}`

extension/intents/nicknames/nicknames.js
ianb: Actually I think it should just be something like:

```js
// A global
let pausedRoutine = null;

...
  async run(context) {
    if (! context.routineExecutor) {
      // error
    }
    pausedRoutine = context.routineExecutor;
    context.routineExecutor.pause();
```

And then later call `pausedRoutine.unpause()` to continue.

Actually, thinking about it a little, I think `unpauseRoutine` should be an exported function from routineExecutor.js, and that module should keep track of any paused routines. Because there's the case when you pause a routine, and another routine is already paused – that actually seems somewhat likely, since it would be easy to forget a routine halfway through. Probably when starting a new routine it makes sense to note the paused routine exists.

In some future version, it probably makes sense to keep a stack of paused routines, but only when routines are called from other routines. Though it's not entirely unreasonable to imagine a routine being paused, and the user then calling another routine during the paused routine to complete the subtask, then resuming the outer routine. But... for now, just showing some messages would be enough.

extension/intents/clipboard/clipboard.js
```diff
@@ -2,7 +2,7 @@ import * as intentRunner from "../../background/intentRunner.js";
 import * as content from "../../background/content.js";
 import * as browserUtil from "../../browserUtil.js";
 
-async function copy(context, copyType, complete = false) {
+async function copy(context, copyType, complete = false, value) {
```
ianb, line 5: Huh, I'm not even entirely sure what a default value in the not-last position does. It would look clearer to me if `value` had a default value too. Or just change the third argument to an options object, since it's hard to keep track of positional arguments with that many.

extension/intents/clipboard/clipboard.toml
```diff
@@ -125,3 +125,9 @@ match = """
 
 [[clipboard.paste.example]]
 phrase = "Paste"
+
+[clipboard.copyValue]
+description = "Copies the value given"
+match = """
+  copy [value]
```
ianb, line 132: This feels like it might catch some things that it's not intended to. Like, "copy the current page's URL" (to think of a phrase we probably don't parse).

I think it should be explicit, like `copy text [value]`

extension/intents/clipboard/contentScript.js
```diff
@@ -65,7 +65,7 @@ this.contentScript = (function() {
 
   communicate.register("copy", message => {
     const { copyType } = message;
-    const result = types[copyType]();
+    const result = types[copyType]() || message.value;
```
ianb, line 68: I'd rather see this special case in `clipboard.js`. Or really `clipboard.copyValue` could just use something besides `copy()` to copy the text. I think you could call `navigator.clipboard.writeText()` directly from the intent handler.

extension/intents/nicknames/nicknames.js
```diff
@@ -114,3 +102,89 @@ intentRunner.registerIntent({
...
+  name: "nicknames.pause",
+  async run(context) {
+    if (context.routineExecutor === undefined) {
+      const exc = new Error("Command not available.");
+      exc.displayMessage = "Command not available.";
+      throw exc;
+    }
+    browser.runtime.sendMessage({
+      type: "presentMessage",
+      message: context.slots.message,
```
ianb, line 116: I'm guessing there should be a default message.

extension/intents/nicknames/nicknames.js
```diff
@@ -114,3 +102,89 @@ intentRunner.registerIntent({
...
+
+    const { name, nextIndex, states, forIndex } = pausedRoutine;
+    const registeredNicknames = await intentRunner.getRegisteredNicknames();
+    const routineExecutor = new RoutineExecutor(
+      registeredNicknames[name].nickname,
+      registeredNicknames[name].contexts,
+      nextIndex,
+      states,
+      forIndex
+    );
```
ianb, line 143: This should definitely all happen in the RoutineExecutor. Pause/continue shouldn't have to know about index, states, or forIndex, these should all stay internal. Then as things are added they won't have to know about those things either.

You could use a `.toJSON()` method to serialize the executor, and `RoutineExecutor.hydrate(json)` to reinstantiate. But I don't think it needs to be kept in storage at all, paused routines can just live in a global variable.

extension/intents/nicknames/nicknames.js
```diff
@@ -114,3 +102,89 @@ intentRunner.registerIntent({
...
+    const objectToLoop = [];
+    if (context.parameters.option === "clipboard") {
+      const result = await navigator.clipboard.readText();
+      const splitResult = result.split("\n");
+      for (const key of splitResult) {
+        objectToLoop.push({ value: key });
+      }
+    }
+
+    if (context.parameters.option === "selected") {
```
ianb, line 161: This should probably be called `selectedTabs` – reading it I thought it meant the text selection.

extension/intents/nicknames/nicknames.js
```diff
@@ -114,3 +102,89 @@ intentRunner.registerIntent({
...
+        objectToLoop.push({ selected: tab.url });
+      }
+    }
+    if (context.parameters.option === "tab") {
+      const tabs = await browser.tabs.query({ currentWindow: true });
+      for (const tab in tabs) {
+        objectToLoop.push({ tab: tab.url });
+      }
+    }
+    if (context.parameters.option === "bookmark") {
```
ianb, line 173: Iterating over bookmarks makes more sense to me if it uses a folder. Like there's always oddball bookmarks in there, but you can put very specific bookmarks in a single folder.

extension/intents/nicknames/nicknames.js
```diff
@@ -114,3 +102,89 @@ intentRunner.registerIntent({
...
+      const splitResult = result.split("\n");
+      for (const key of splitResult) {
+        objectToLoop.push({ value: key });
+      }
+    }
+
+    if (context.parameters.option === "selected") {
+      const selectedTabs = await browser.tabs.query({ highlighted: true });
+      for (const tab of selectedTabs) {
+        objectToLoop.push({ selected: tab.url });
```
ianb, line 164: Using strings here for now is probably fine, but ultimately there should be special wrappers around each type. For instance, you might want to do:

```
for each selected tab x
activate tab x
pause
close tab x
end for
```

For some ways of using the tab the URL is appropriate, for others the tab object or tab ID is appropriate.

I think it's best _not_ to use the actual raw tab object, because then consumers have to test a bunch of types to figure things out. But instead it would be a Tab object with a `.toString() {return this._tab.url}` method, and maybe some other methods specific to tabs, or it inherits from an abstract base class that implements tab-related fallbacks for non-tab objects.

But this can be a followup.

extension/intents/nicknames/nicknames.js
```diff
@@ -114,3 +102,89 @@ intentRunner.registerIntent({
...
+      }
+    }
+    if (context.parameters.option === "tab") {
```
ianb, line 167: Instead of `.option` I'd call this `.dataSource`

extension/intents/nicknames/nicknames.toml
```diff
@@ -60,4 +60,33 @@ match = """
...
+  for each value in (clipboard) [option=clipboard]
+  for each tab [option=tab]
+  for each bookmark [option=bookmark]
+  for each selected (tab |) [option=selected]
+"""
+
+[nicknames.endForLoop]
+description = "Statement to end loop"
+match = """
+  endfor
```
ianb, line 90: Should also match `end for`

extension/intents/nicknames/nicknames.toml
```diff
@@ -60,4 +60,33 @@ match = """
...
+  continue routine
+"""
+
+[nicknames.startForLoop]
+description = "Statement to start loop"
+match = """
+  for each value in (clipboard) [option=clipboard]
+  for each tab [option=tab]
+  for each bookmark [option=bookmark]
+  for each selected (tab |) [option=selected]
```
ianb, line 84: I think `tab` should be required in this phrase

extension/intents/nicknames/routineExecutor.js
```diff
@@ -0,0 +1,102 @@
+/* globals log */
+
+import * as intentRunner from "../../background/intentRunner.js";
+
+export class RoutineExecutor {
+  constructor(
+    routineName,
+    subcommands,
+    index = 0,
+    states = [{}],
```
ianb, line 10: These should probably start out as null, as it makes sense to treat not-in-a-for-loop as separate from being a in for loop.

extension/intents/nicknames/routineExecutor.js
```diff
@@ -0,0 +1,102 @@
...
+    let hadError = false;
+    let errorMessage = null;
+
+    log.info(
+      "  Running subintent",
+      subcommand,
+      subcommand.name,
+      subcommand.slots,
+      this.index
+    );
```
ianb, line 40: The log message should probably go after `this.mapState()` so it shows the substitution.

extension/intents/nicknames/nicknames.toml
```diff
@@ -60,4 +60,33 @@ match = """
...
+[nicknames.continue]
+description = "Continues a routine"
+match = """
+  continue routine
+"""
+
+[nicknames.startForLoop]
+description = "Statement to start loop"
+match = """
+  for each value in (clipboard) [option=clipboard]
```
ianb, line 81: It took me a bit to realize there's an implied for loop variable, `value` (or `bookmark` etc).

That does seem reasonable, though I think `for each selected tab` should still assign `tab`.

But I think it would be good to also allow variable names. Specifically something like `tab` could easily get eaten up by the parser. E.g., `find tab` may not parse with "tab" in a slot (it actually might depending on the phrase matcher... but it's hard to be sure).

I'd suggest the phrasing `for each value in clipboard (as | named | as name) [variable] [dataSource=clipboard]`. This would require duplicating each of these phrases here.

extension/intents/nicknames/routineExecutor.js
```diff
@@ -0,0 +1,102 @@
...
+        `Routine ${this.routineName} encountered error ${errorMessage} while running ${subcommand.utterance}`
+      );
+      return true;
+    }
+
+    return this._stop === true;
+  }
+
+  startLoop(states) {
+    this.states = states;
```
ianb, line 60: This doesn't allow for nested loops. I'm pretty sure nested loops are going to be a bit too complex for what we're doing, but as implemented it actually _allows_ nested loops, they just won't work right.

To make it clearer here, I'd set `this.states` to null when not in a for loop, and check here that it's null and otherwise throw an error (instead of overwriting the state with no error).

extension/intents/nicknames/routineExecutor.js
```diff
@@ -0,0 +1,102 @@
...
+    this.forIndex = this.index;
+  }
+
+  endLoop() {
+    this.states.shift();
+    if (this.states.length > 0) {
+      this.index = this.forIndex;
```
ianb, line 67: I had to read through to here to understand quite what `forIndex` was. Maybe a better name for both forIndex and `this.index` would be [program counter](https://en.wikipedia.org/wiki/Program_counter). It makes it clearer what it's used for.

extension/intents/nicknames/nicknames.js
ianb: An example of how to create a tab object that  an be used specially for some intents, but treated like a URL by default:

```js
class TabVariable {
  constructor(tabObj) {
    this.tabObj = tabObj;
  }
  toString() {
    return this.tabObj.url;
  }
  closeTab() {
    return browser.tabs.remove(this.tabObj.id);
  }
}

registerIntent("tabs.closeTab", async (context) => {
  if (context.slots.query.closeTab) {
    await context.slots.query.closeTab();
  }
});
```

extension/intents/nicknames/routineExecutor.js
```diff
@@ -0,0 +1,118 @@
...
+      this.states = null
+    }
+  }
+
+  async run() {
+    if (pausedRoutineExecutor !== null) {
+      const exc = new Error("Another routine is already active.");
+      exc.displayMessage = "Another routine is already active.";
+      throw exc;
+    }
```
ianb, line 87: We should make a followup issue for nested routines, as it seems like a natural thing to do.

extension/intents/nicknames/routineExecutor.js
```diff
@@ -0,0 +1,118 @@
...
+      const exc = new Error("Encountered nested loops. Not yet supported.");
+      exc.displayMessage = "Encountered nested loops. Not yet supported.";
+      throw exc;
+    }
+    this.states = states;
+    this.startLoopProgramCounter = this.programCounter;
+  }
+
+  endLoop() {
+    this.states.shift();
```
ianb, line 74: I think both here and in the intent for ending loops there's nothing that will stop you from ending a loop when there's no active loop. So you'll probably get some kind of exception?

extension/options/optionsController.jsx
```diff
@@ -206,7 +205,7 @@ export const OptionsController = function() {
     }
 
     setRegisteredNicknames(registeredNicknames);
-    return true;
+    return {allowed: true};
```
ianb, line 208: This change seems odd...?

ianb:
I think it's ready to land, amid the conflicts it looks like there's some prettier errors but I'll fix them in master.

PR #1807, opened 2020-07-12
Fixes #1789 Use free text search in tabs.js:getMatchingTabs()

### LINKED ISSUE #1789: Use free text search in tabs.js:getMatchingTabs()
The function `getMatchingTabs()` basically does a string search of the URL. That's not a great way to match, especially with voice input.

Instead we should get all tabs that otherwise match the query, and use `query` as a text search of the URL and title of the tabs (it doesn't need to include tab content).

AbhiVaidya95:

- [✗] Run `npm test` on your machine
- [✓] Run `npm run format` to keep code formatted consistently
- [✓] Reference the issue you are fixing in at least one of your commit messages, as `Fixes #X`

### DIFF
--- a/extension/intents/tabs/tabs.js
+++ b/extension/intents/tabs/tabs.js
@@ -1,4 +1,4 @@
-/* globals buildSettings */
+/* globals buildSettings,Fuse,log */
 import * as intentRunner from "../../background/intentRunner.js";
 import English from "../../language/langs/english.js";
 import * as browserUtil from "../../browserUtil.js";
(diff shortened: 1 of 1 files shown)


extension/intents/tabs/tabs.js
```diff
@@ -469,13 +469,58 @@ async function getMatchingTabs(options) {
...
+    // use Fuse.js to parse the most probable response?
+    const fuse = new Fuse(combinedTabContent, fuseOptions);
+    const matches = fuse.search(options.query);
+    log.debug("find matches:", matches);
+    // TODO account for multiple matches
+    if (!matches.length) {
+      const e = new Error("No matching tab found");
+      e.displayMessage = "No matching tab found";
+      throw e;
+    }
```
ianb, line 518: Generally we do this kind of error checking in the intent handler itself. Since this is a generic function that returns a list of tabs, it doesn't need to do error checking, simply returning an empty list is OK.

extension/intents/tabs/tabs.js
```diff
@@ -469,13 +469,58 @@ async function getMatchingTabs(options) {
...
+      const result = {
+        tabId: tab.id,
+        title: tab.title,
+        url: tab.url,
+      };
+
+      combinedTabContent.push(result);
+    }
+
+    combinedTabContent = combinedTabContent.flat();
```
ianb, line 507: This probably isn't necessary? It looks like the array should be flat already.

AbhiVaidya95: @ianb Thanks for the suggestion, I have removed the unnecessary code.

PR #1813, opened 2020-07-15
 Fixes #1484 Create pageName type

### LINKED ISSUE #1484: Create namedPage type
This is a followup to #1302:

Right now there's some special handling of named pages, but instead we'd like to put "named page" into other phrases. For instance `open [name:namedPage]` that would match named pages.

To do this we need a new page type (in `entityTypes.js`) that is dynamically filled with the available names. This would also mean that page names would respect aliases and ignore stopwords.

I believe `textMatching.Alternatives` can accept a mutable array, and if that array is changed then the pattern matching will change. So we can instantiate the alternatives, and modify it when storage is initialized and as commands are executed that change the named pages. Ideally this would happen in the functions suggested by #1483

Simpcyclassy:

### DIFF
--- a/extension/background/intentRunner.js
+++ b/extension/background/intentRunner.js
@@ -6,10 +6,10 @@ import { metadata } from "../intents/metadata.js";
 import { compile, splitPhraseLines } from "../language/compiler.js";
 import { PhraseSet } from "../language/findMatch.js";
 import * as settings from "../settings.js";
+import { registerHandler, sendMessage } from "./communicate.js";
 import { entityTypes } from "./entityTypes.js";
 import * as intentParser from "./intentParser.js";
 import * as telemetry from "./telemetry.js";
-import { registerHandler, sendMessage } from "./communicate.js";
 
 const FEEDBACK_INTENT_TIME_LIMIT = 1000 * 60 * 60 * 24; // 24 hours
 // Only keep this many previous intents:
@@ -541,16 +541,18 @@ export async function getRegisteredPageName(name) {
   const result = await browser.storage.sync.get("pageNames");
   pageNames = result.pageNames;
 
+  return pageNames;
+}
+
+export async function unregisterPageName(name) {
+  pageNames = getRegisteredPageName(name);
   if (!pageNames[name] || !name) {
     const exc = new Error("No page name to remove");
     exc.displayMessage = `The page name "${name}" not found`;
     throw exc;
   }
-  return pageNames;
-}
-
-export async function unregisterPageName(name) {
   delete pageNames[name];
+
   log.info("Removed nickname for page", name);
   await browser.storage.sync.set({ pageNames });
 }
--- a/extension/intents/navigation/navigation.js
+++ b/extension/intents/navigation/navigation.js
@@ -1,13 +1,13 @@
+import { sendMessage } from "../../background/communicate.js";
+import * as content from "../../background/content.js";
 import * as intentRunner from "../../background/intentRunner.js";
-import * as serviceList from "../../background/serviceList.js";
 import * as languages from "../../background/languages.js";
 import * as pageMetadata from "../../background/pageMetadata.js";
-import * as searching from "../../searching.js";
-import * as content from "../../background/content.js";
+import * as serviceList from "../../background/serviceList.js";
 import * as browserUtil from "../../browserUtil.js";
+import * as searching from "../../searching.js";
 import { metadata } from "../../services/metadata.js";
 import { performSearchPage } from "../search/search.js";
-import { sendMessage } from "../../background/communicate.js";
 
 const QUERY_DATABASE_EXPIRATION = 1000 * 60 * 60 * 24 * 30; // 30 days
 const queryDatabase = new Map();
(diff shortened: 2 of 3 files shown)


extension/background/entityTypes.js
```diff
@@ -21,9 +21,19 @@ for (const id in metadata.music) {
...
 
+export const namedPages = [];
+
+for (const id in metadata.pageNames) {
+  const item = metadata.pageNames[id];
+  for (const name of item.names) {
+    namedPages.push(name);
+  }
+}
+
```
Simpcyclassy, line 32: I understand I need to create a `new Alternative` here can I get a bit more clarity on that @ianb

extension/background/intentRunner.js
```diff
@@ -525,32 +525,27 @@ registerHandler("getRegisteredNicknames", getRegisteredNicknames);
...
 let pageNames = {};
 
+export async function getRegisteredPageName() {
+  const result = await browser.storage.sync.get("pageNames");
+  pageNames = result.pageNames;
+
+  return pageNames;
+}
+
 export async function registerPageName(name, { url }) {
```
Simpcyclassy, line 535: I assume this is the function where I am to save the pageNames to the entityTypes, it's not clear to me how the services communicate

extension/background/entityTypes.js
```diff
@@ -21,9 +21,19 @@ for (const id in metadata.music) {
...
 export const entityTypes = convertEntities({
   serviceName: allServiceNames,
   musicServiceName: musicServiceNames,
+  name: namedPages,
   lang: languageNames(),
   smallNumber: English.numberNames,
 });
```
ianb, line 39: convertEntities does its own thing. It might actually be OK, but then if you look at `entityTypes.name` (it would be better to call it `.pageName`) you'll see it is an instance of Alternatives.

A good approach here is to use things like `console.log(entityTypes.name)` to see what's what.

But to add a new thing to the entity you'd do something like:

```js
export function addPageName(name) {
  entityTypes.pageNames.alternatives.push(makeWordMatcher(name));
}
```

(Note makeWordMatcher isn't exported at the moment, but could be)

extension/background/intentRunner.js
ianb: Yes, this should call a new function in `entityTypes.js`

extension/intents/navigation/navigation.js
```diff
@@ -34,32 +34,31 @@ function saveTabQueryToDatabase(query, tab, url) {
 intentRunner.registerIntent({
   name: "navigation.navigate",
   async run(context) {
-    const query = context.slots.query.toLowerCase();
-    const result = await browser.storage.sync.get("pageNames");
-    const pageNames = result.pageNames;
+    const name = context.slots.name.toLowerCase();
```
ianb, line 37: I don't think these changes to navigation.js are necessary or related to the rest of the changes

extension/background/intentRunner.js
```diff
@@ -525,32 +525,27 @@ registerHandler("getRegisteredNicknames", getRegisteredNicknames);
...
   name = name.toLowerCase();
   const creationDate = Date.now();
-  const result = await browser.storage.sync.get("pageNames");
-  pageNames = result.pageNames || {};
+  pageNames = getRegisteredPageName() || {};
```
ianb, line 538: We probably won't need this any longer. Instead we can create a new intent that just matches the phrase:

```
[name:pageName]
```

And the new intent should look up and open the page.

extension/background/intentRunner.js
```diff
@@ -523,34 +523,31 @@ export function getRegisteredNicknames() {
...
+
+  return pageNames;
+}
 
 export async function registerPageName(name, { url }) {
   name = name.toLowerCase();
   const creationDate = Date.now();
-  const result = await browser.storage.sync.get("pageNames");
-  pageNames = result.pageNames || {};
+  const pageNames = getRegisteredPageName() || {};
```
ianb, line 538: Missing await?

extension/background/intentRunner.js
```diff
@@ -523,34 +523,31 @@ export function getRegisteredNicknames() {
...
-  if (!pageNames[name] || !name) {
-    const exc = new Error("No page name to remove");
-    exc.displayMessage = `The page name "${name}" not found`;
-    throw exc;
-  }
-  return pageNames;
-}
-
 export async function unregisterPageName(name) {
+  const pageNames = getRegisteredPageName();
```
ianb, line 548: also needs await

extension/background/intentRunner.js
```diff
@@ -548,33 +548,28 @@ export function getRegisteredRoutines() {
 
 registerHandler("getRegisteredRoutines", getRegisteredRoutines);
 
-let pageNames = {};
+export async function getRegisteredPageName() {
+  const result = await browser.storage.sync.get("pageNames");
+  let pageNames = result.pageNames;
+  pageNames = pageName;
```
ianb, line 554: This line is definitely wrong

extension/intents/routines/routines.js
```diff
@@ -98,7 +98,13 @@ intentRunner.registerIntent({
   name: "routines.removePageName",
   async run(context) {
     const name = context.slots.name;
-    await intentRunner.getRegisteredPageName(name);
+    const pageNames = await intentRunner.getRegisteredPageName(name);
```
ianb, line 101: Here `getRegisteredPageName()` is used with an argument, but the function as defined doesn't take any arguments.

Really there's a confusion here, as `getRegisteredPageName()` is _also_ used elsewhere without an argument (in `registerPageName()`). But in that case it's being used incorrectly.

Be sure when you are changing a function that you look for everywhere where the function is used, to make sure you are changing it in a way that's compatible with how it's being used.

PR #1821, opened 2020-07-18
Fixes #1771 Intent-finding intent Fixed

### LINKED ISSUE #1771: Intent-finding intent
Not sure about the phrase, but something like "help about tabs" or "commands for tabs", and it would show examples from intents where the example or intent description included "tab".

AbhiVaidya95:

- [✘] Run `npm test` on your machine
- [✓] Run `npm run format` to keep code formatted consistently
- [✓] Reference the issue you are fixing in at least one of your commit messages, as `Fixes #X`

### DIFF
--- a/extension/intents/self/contentScript.js
+++ b/extension/intents/self/contentScript.js
@@ -1,0 +1,45 @@
+/* globals communicate, Fuse */
+
+communicate.register("focusField", async message => {
+  const helpElement = message.query;
+  const elements = [];
+  const searchContent = [];
+  for (const el of document.querySelectorAll(".intent,.description")) {
+    elements.push(el);
+    searchContent.push({
+      element: el,
+      innerText: el.innerText,
+    });
+  }
+  const options = {
+    shouldSort: true,
+    tokenize: true,
+    matchAllTokens: true,
+    findAllMatches: true,
+    includeScore: true,
+    threshold: 0.3,
+    location: 0,
+    distance: 100,
+    maxPatternLength: 32,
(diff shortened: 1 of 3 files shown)


extension/intents/self/contentScript.js
```diff
@@ -0,0 +1,45 @@
+/* globals communicate, Fuse */
+
+communicate.register("focusField", async message => {
+  const helpElement = message.query;
+  const elements = [];
+  const searchContent = [];
+  for (const el of document.querySelectorAll(".intent,.description")) {
```
ianb, line 7: This should include `.utterance` too – it will match something like "scroll" then (which is in a `<div class=utterance>`)

extension/intents/self/self.toml
```diff
@@ -113,3 +113,13 @@ match = """
 
 [[self.simpleTest.example]]
 phrase = "test"
+
+[self.helpIntent]
+description = "This will try to find help for for the intent on lexicon page"
+match = """
+  (show | open |) help (for | on |) [query] (to me | for me |)
+  (show | open |) command (for | on |) [query] (to me | for me |)
```
ianb, line 121: I think this would be a better phrase:

```
(show | open |) (help | command{s}) (for | on | about |) [query] (to | for |) (me |)
```

It should match a few other forms of the phrase.

ianb:
Actually I'm going to quick land this and make those changes, because I want to land #1815 and I want to get all the conflicts out of the way

PR #1837, opened 2020-07-23
Fixes #1836 added timer functionalities for pomodoro

### LINKED ISSUE #1836: Pomodoro routine
Timer changes to allow for a pomodoro routine; In a loop, start a timer, wait for timer to finish, start a timer for break; repeat this 4 times;

danielamormocea:
Fixes #1836 

A routine for pomodoro (with 4 breaks) can now be written as:

`
for each number from 1 to 4

start timer for 25 minutes

wait for timer to end

start timer for 3 minutes

wait for timer to end

endfor
`

### DIFF
--- a/extension/intents/timer/timer.js
+++ b/extension/intents/timer/timer.js
@@ -23,7 +23,7 @@ class TimerController {
         this.activeTimer.close();
         this.activeTimer = null;
         resolve(true);
-      }
+      };
     });
   }
 
@@ -32,9 +32,9 @@ class TimerController {
     return this.activeTimer;
   }
 
-  wait = async () => {
+  async waitTimer() {
     return this.waitPromise;
-  } 
+  };
 }
 
 class Timer {
@@ -104,7 +104,6 @@ class Timer {
       if (result) {
         return;
       }
-      
     } catch (e) {
       catcher.capture(e);
     }
@@ -240,7 +239,6 @@ registerHandler("timerAction", message => {
   return timerController[message.method](...(message.args || []));
 });
 
-
 intentRunner.registerIntent({
   name: "timer.wait",
   async run() {
@@ -251,6 +249,5 @@ intentRunner.registerIntent({
       throw e;
     }
     await timerController.wait();
-
-  }
+  },
 });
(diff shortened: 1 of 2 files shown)


extension/intents/timer/timer.js
```diff
@@ -13,21 +14,27 @@ class TimerController {
...
     this.activeTimer = new Timer(totalInMS, context);
     this.lastActiveTimer = { ...this.activeTimer };
     this.activeTimer.start();
+    this.waitPromise = new Promise((resolve, reject) => {
+      this.closeActiveTimer = () => {
+        this.activeTimer.close();
+        this.activeTimer = null;
+        resolve(true);
+      };
+    });
```
ianb, line 27: You could also use `util.makeNakedPromise()` here to make the logic slightly simpler (then `closeActiveTimer()` could just call `this.waitPromise.resolve(true)`).

ianb: Is closeActiveTimer really going to work? It seems like it's only called by `"timer.close"`, and not by the timer actually running out...? As far as I can tell, when the timer runs out it opens the popup but doesn't call this close method.

extension/intents/timer/timer.js
```diff
@@ -232,3 +238,16 @@ intentRunner.registerIntent({
...
+intentRunner.registerIntent({
+  name: "timer.wait",
+  async run() {
+    const activeTimer = timerController.getActiveTimer();
+    if (activeTimer === null || activeTimer.paused === false) {
+      const e = new Error("Failed to wait for timer to finish.");
+      e.displayMessage = "No active timer.";
+      throw e;
+    }
+    await timerController.wait();
```
ianb, line 251: This should be waitTimer

extension/intents/timer/timer.js
```diff
@@ -13,21 +14,27 @@ class TimerController {
...
   }
 
   restoreTimer(context) {
     this.setActiveTimer(this.lastActiveTimer.totalInMS, context);
     return this.activeTimer;
   }
+
+  async waitTimer() {
+    return this.waitPromise;
+  };
```
ianb, line 37: semicolon?

extension/intents/timer/timer.js
ianb: I am worried that an intent that lasts for 15 minutes is... fragile. I'd rather this only work in a routine (it only makes sense in a routine), and that it pause the routineexecutor and resume it in a certain amount of time.

PR #1846, opened 2020-07-25
Fixes #1820 added onboarding views for wakeword

### LINKED ISSUE #1820: Wakeword onboarding
During onboarding we should provide a choice for triggers, either wakeword or manual (mic icon click, or keyboard shortcut). See attached mockup for new onboarding step.

- New first step to onboarding provides the option to enable wakeword.
- Based on wakeword choice text changes will be necessary for step 3 and 4.
-- If wakeword has been enabled the highlighted text should read "The microphone is only active when triggered using “hey Firefox”,  a button press or keyboard shortcut."
- If wakeword has been enabled the startup educating about how to trigger should change (see mockup)

wakeword-onboarding

danielamormocea:
Fixes #1820 

Added onboarding modal and page for wakeword settings;

### DIFF
--- a/extension/onboarding/onboarding.css
+++ b/extension/onboarding/onboarding.css
@@ -200,8 +200,7 @@ p {
   background-image: linear-gradient(to right, #6173ff, #f10366, #ff9100);
 }
 
-#wakeword-try-it h1:before,
-#get-started h1:before {
+.mic-icon h1:before {
   background-image: url("../popup/images/mic.svg");
   background-size: 40px auto;
 }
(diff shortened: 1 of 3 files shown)


extension/onboarding/onboardingView.jsx
```diff
@@ -144,8 +193,9 @@ const OptinAudioDescription = () => {
         collection.
       </p>
       <p className="warning">
-        The microphone is only active when triggered with a button press or
-        keyboard shortcut.
+        {wakewordActive
+          ? 'The microphone is only active when triggered using "Hey Firefox", a button press or keyboard shortcut.'
```
ianb, line 197: Well... this isn't really true, with wakewordActive the microphone will always be on, but won't be sent to any server.

@awallin do you have a text suggestion here? Maybe "With Hey Firefox active the microphone will be listening, but audio will not be sent to a server until after you say Hey Firefox or activate with a button press or keyboard shortcut."

extension/onboarding/onboarding.css
```diff
@@ -212,10 +221,12 @@ p {
   background-image: url("./images/icon-superhero.svg");
 }
 
+#wakeword-try-it p,
```
ianb, line 224: It would be good to change `try-it` to a class instead of an id, so there doesn't have to be two selectors on each of these.

awallin:
Let's go with...

> Using “Hey Firefox” the microphone is but audio will not be collected or sent to a server until after you say “Hey Firefox” or activate using a button press or keyboard shortcut.

PR #1853, opened 2020-07-28
Fix feedback display for timer

Gwe-n:

The initial problem is just that the `onSubmitFeedback` binding in `IntentFeedback` is missing. I added it. 

There was another problem as well. When you get feedback on a first search, `lastIntent` doesn't exist. 

So I added an optional chaining on this object, and I considered that if it didn't exist you could approximate that `minutesAgo` is equal to 1.

I also noticed that when you write a feedback for the first time (when you did a single search, without doing a "next" or "previous" or when you run the timer the first time) the popup closed while you were writing the comment. 

feedback-closed

So I removed the call to `runFollowup()` in the "switch/case" `handleFollowup` in the `popupController`.

However, it may be better to deal with this issue in a separate commit. 
I was just afraid that the problem with the timer feedback feature might not be solved properly because the popup would close while writing.
But in my opinion these are two different problems. 
Let me know, I'll change it.


ianb:
Thanks for looking into this. I think removing `runFollowup()` probably isn't the right thing. Instead we should cancel the close when feedback starts. I'll fix that up in a followup

PR #1874, opened 2020-08-04
Fixes #1838 stop routine intent

### LINKED ISSUE #1838: One should be able to stop a routine
Now that pause routine is a thing, one should be able to also stop a routine;
Create a 'stop routine' intent

danielamormocea:
Fixes #1838

added intent "stop routine" to be able to cancel a routine (most preferably when it's paused);

### DIFF
--- a/extension/intents/routines/routineExecutor.js
+++ b/extension/intents/routines/routineExecutor.js
@@ -3,19 +3,22 @@
 import * as intentRunner from "../../background/intentRunner.js";
 
 export let pausedRoutineExecutor = null;
+export let currentRoutineExecutor = null;
 export class RoutineExecutor {
   constructor(
     routineName,
     subcommands,
     programCounter = 0,
     states = null,
-    startLoopProgramCounter = 0
+    startLoopProgramCounter = 0,
+    stopRoutine = false
   ) {
     this.routineName = routineName;
     this.subcommands = subcommands;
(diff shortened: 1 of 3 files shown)


extension/intents/routines/routineExecutor.js
```diff
@@ -3,19 +3,22 @@
 import * as intentRunner from "../../background/intentRunner.js";
 
 export let pausedRoutineExecutor = null;
+export let currentRoutineExecutor = null;
```
ianb, line 6: Huh, I didn't realize you could export a non-constant value. And from what I'm reading maybe it live updates when imported? Though that's a bit surprising to me that it works.

I'd usually use a getter function, but I suppose if this works...

extension/intents/routines/routineExecutor.js
```diff
@@ -85,14 +88,19 @@ export class RoutineExecutor {
       exc.displayMessage = "Another routine is already active.";
       throw exc;
     }
+    currentRoutineExecutor = this;
     for (
       ;
       this.programCounter < this.subcommands.length;
       this.programCounter++
     ) {
+      if (this.stopRoutine === true) {
```
ianb, line 97: Unless `stopRoutine` can be something besides true/false, then it would be best to just say `if (this.stopRoutine)`

PR #1894, opened 2020-08-06
Fixes #1800 add speech telemetry

### LINKED ISSUE #1800: Add speech-related parameters to Telemetry
We should add both information about speech output, and some of the metadata that we are now collecting as part of the speech text extraction:

- [ ] A boolean if something was spoken
- [ ] A boolean if the speech preference is on
- [ ] The length of text extracted for speech
- [ ] The kind of card (string) that was found

The last item is something we would find interesting regardless of speech output, as it would give us some information about the many different kinds of fallback search results people are getting.

jcambre:


ianb:
So this will require:

- speechOutputEnabled (boolean)
- outputSpoken (boolean)
- cardType (string)
- outputSpokenWords (number)

extension/popup/popupController.jsx
```diff
@@ -337,6 +337,10 @@ export const PopupController = function() {
     utterance.onend = () => {
       closePopup(2000);
```
ianb, line 338: Let's add outputSpokenCompleted (true) to this method

jcambre:
We should also include speech input locale in telemetry

